PR #8, opened 2016-06-07
Update to 0.9.0rc0

danielfrg:
Update to `0.9.0rc0` now they released linux and python 3.4 wheel so its now supported :)

I added mysellf as maintainer hopefully thats ok.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -4,16 +4,19 @@
 
 if [ `uname` == Darwin ]; then
     if [ "$PY_VER" == "2.7" ]; then
-        pip install --no-deps https://storage.googleapis.com/tensorflow/mac/tensorflow-0.8.0-py2-none-any.whl
+        pip install --no-deps https://storage.googleapis.com/tensorflow/mac/tensorflow-0.9.0rc0-py2-none-any.whl
     else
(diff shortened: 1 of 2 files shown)


jakirkham:
See this discussion about rc and dev releases. ( https://github.com/conda-forge/conda-forge.github.io/issues/147 ) Unfortunately, this is not settled.

danielfrg:
Thanks for the link to the discussion @jakirkham. I'll let this open for now and update to `0.9.0` when its released finally.

jakirkham:
FWIW, I think everyone wants to see this. So, it isn't a question of whether, but how this happens. Also, there is a related question of how one can control which packages use dev/rc releases.

jjhelmus:
@danielfrg This looks good, but I agree that we should wait until `0.9.0` is relased to create a conda package.  

Related I see in the **Major Features and Improvements** section of the release a point on "Added makefile for better cross-platform build support (C API only)".  Have you examined this at all?  Will it allow us to build the package from source without needing to install Bazel?

jjhelmus:
Sorry to answer my own question, but it looks like the make is only for creating a static C++ tensorflow library not for the Python package, [details here](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/makefile).

jjhelmus:
Closing now that #9 has been merged which updates to 0.9.0.

Issue #6, opened 2016-05-20
Build tensorflow from source not whl files.


jjhelmus:
Currently the tensorflow recipe is repackaging the whl files provided by Google into conda packages.  The whl files link against a newer version of glibc which limited what Linux distribution that the conda package works on and are only available for Python 2.7 and 3.4 on Linux.

A "better" solution would be to build the package from source, but this requires installing [Bazel](http://bazel.io/) which [requires Java JDK 8](http://bazel.io/docs/install.html).  I have successfully built a conda package for 0.8.0 from source using a [Ubunutu 14.04 Docker container](https://github.com/jjhelmus/wip_conda_recipes/tree/master/tensorflow) with Java+Bazel.  We could do the same here but it would require quite a bit of customization of this feedstock.  Anyone have opinion on this topic?

rgommers:
There's 3 people reporting at https://github.com/numpy/numpy/issues/8076 that installing ``tensorflow`` breaks numpy, and it looks related to this wheel repackaging business.

rgommers:
[This comment](https://github.com/scipy/scipy.org/pull/179#issuecomment-247179912) says that indeed tensorflow pulls in a numpy wheel somehow. So that's pretty badly broken.

jakirkham:
Ouch! Yeah the build situation here is not great. It is unfortunately quite difficult to build `tensorflow` from source. That said, @jjhelmus added `bazel` recently (merged today), which should help.

hajapy:
I'm a little confused, our build recipe installs tensorflow wheel via pip with --no-deps, so the only numpy should be one specified by the meta.yaml run time requirements.

rgommers:
I also don't really understand it, but it's reported by 4 people so should be reproducible.

jakirkham:
Of course, I think it is more surprise at the issue's existence than any doubt that people are running into an issue. Though you make a good point, steps to reproduce the issue would be very helpful in debugging it. Was that in the issue and I simply missed it?

rgommers:
No, they all said "installed tensorflow with conda, then numpy broke". Asked for exact install commands.

jakirkham:
Great, thanks Ralf.

pgunn:
Is it possible to copy the build instructions from the non-conda-forge builds of tensorflow? Those versions don't seem to have the GLIBC version dependency (tested on RHEL7, which has a very dated glibc).

jakirkham:
[...]

PR #12, opened 2016-08-24
Build tensorflow from source on Linux

jjhelmus:
Build tensorflow from source on Linux after building Bazel from source.
Requires the install of java and a few other packages using yum.

### DIFF
--- a/ci_support/run_docker_build.sh
+++ b/ci_support/run_docker_build.sh
@@ -50,17 +50,17 @@ yum install -y java-1.8.0-openjdk-devel unzip zip zlib-devel rsync
 
 
 # Embarking on 3 case(s).
-    set -x
-    export CONDA_PY=27
-    set +x
-    conda build /recipe_root --quiet || exit 1
-    /feedstock_root/ci_support/upload_or_check_non_existence.py /recipe_root conda-forge --channel=main || exit 1
+#    set -x
+#    export CONDA_PY=27
+#    set +x
+#    conda build /recipe_root --quiet || exit 1
+#    /feedstock_root/ci_support/upload_or_check_non_existence.py /recipe_root conda-forge --channel=main || exit 1
 
-    set -x
-    export CONDA_PY=34
-    set +x
-    conda build /recipe_root --quiet || exit 1
-    /feedstock_root/ci_support/upload_or_check_non_existence.py /recipe_root conda-forge --channel=main || exit 1
+#    set -x
+#    export CONDA_PY=34
+#    set +x
+#    conda build /recipe_root --quiet || exit 1
+#    /feedstock_root/ci_support/upload_or_check_non_existence.py /recipe_root conda-forge --channel=main || exit 1
 
     set -x
     export CONDA_PY=35
--- a/ci_support/run_docker_build.sh
+++ b/ci_support/run_docker_build.sh
@@ -14,7 +14,6 @@ config=$(cat <<CONDARC
 
 channels:
  - conda-forge
-
  - defaults # As we need conda-build
 
 conda-build:
@@ -39,9 +38,16 @@ echo "$config" > ~/.condarc
 # A lock sometimes occurs with incomplete builds. The lock file is stored in build_artefacts.
 conda clean --lock
 
-conda update --yes --all
-conda install --yes conda-build
-conda info
+conda install --yes --quiet conda-forge-build-setup
+source run_conda_forge_build_setup
+
+
+# Install the yum requirements defined canonically in the
+# "recipe/yum_requirements.txt" file. After updating that file,
+# run "conda smithy rerender" and this line be updated
+# automatically.
+yum install -y java-1.8.0-openjdk-devel unzip zip zlib-devel rsync
+
 
 # Embarking on 3 case(s).
     set -x
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -16,14 +16,10 @@ if [ `uname` == Linux ]; then
     curl -s -L -O https://github.com/bazelbuild/bazel/archive/0.3.1.tar.gz
     tar xf 0.3.1.tar.gz
     cd bazel-0.3.1
-    #./compile.sh
-    #mkdir -p ~/bin
-    #cp output/bazel ~/bin/
+    ./compile.sh
+    mkdir -p ~/bin
+    cp output/bazel ~/bin/
     cd ..
-
-    # debug
-    pwd
-    ls
 
     # Compile tensorflow from source
 
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -34,7 +34,7 @@ if [ `uname` == Linux ]; then
     cat tensorflow/tools/swig/swig.sh
 
     # build wheel using bazel
-    bazel build -c opt //tensorflow/tools/pip_package:build_pip_package
+    bazel build --jobs 2 -c opt //tensorflow/tools/pip_package:build_pip_package
     mkdir $SRC_DIR/tensorflow_pkg
     bazel-bin/tensorflow/tools/pip_package/build_pip_package $SRC_DIR/tensorflow_pkg
 
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -16,10 +16,14 @@ if [ `uname` == Linux ]; then
     curl -s -L -O https://github.com/bazelbuild/bazel/archive/0.3.1.tar.gz
     tar xf 0.3.1.tar.gz
     cd bazel-0.3.1
-    ./compile.sh
-    mkdir -p ~/bin
-    cp output/bazel ~/bin/
+    #./compile.sh
+    #mkdir -p ~/bin
+    #cp output/bazel ~/bin/
     cd ..
+
+    # debug
+    pwd
+    ls
 
     # Compile tensorflow from source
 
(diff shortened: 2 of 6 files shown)


jjhelmus:
This address the Linux side of #6, not sure how do deal with the OS X side or if it is required.

ci_support/run_docker_build.sh
```diff
@@ -39,9 +38,16 @@ echo "$config" > ~/.condarc
...
-conda info
+conda install --yes --quiet conda-forge-build-setup
+source run_conda_forge_build_setup
+
+
+# Install the yum requirements defined canonically in the
+# "recipe/yum_requirements.txt" file. After updating that file,
+# run "conda smithy rerender" and this line be updated
+# automatically.
+yum install -y java-1.8.0-openjdk-devel unzip zip zlib-devel rsync
```
ocefpaf: With `yum_requirements.txt` you don't really need this here.

recipe/yum_requirements.txt
```diff
@@ -0,0 +1,5 @@
+java-1.8.0-openjdk-devel
+unzip
+zip
+zlib-devel
```
ocefpaf: No go with the packaged `zlib`?

recipe/build.sh
```diff
@@ -11,11 +11,34 @@ if [ `uname` == Darwin ]; then
...
 if [ `uname` == Linux ]; then
-    if [ "$PY_VER" == "2.7" ]; then
-        pip install --no-deps https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.9.0-cp27-none-linux_x86_64.whl
-    elif [ "$PY_VER" == "3.4" ]; then
-        pip install --no-deps https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.9.0-cp34-cp34m-linux_x86_64.whl
-    elif [ "$PY_VER" == "3.5" ]; then
-        pip install --no-deps https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-0.9.0-cp35-cp35m-linux_x86_64.whl
-    fi
+
+    # Install Bazel from source
```
ocefpaf: Any chance for a `bazel` package?

ci_support/run_docker_build.sh
jjhelmus: That line was added by conda smithy, I think it is needed.

recipe/yum_requirements.txt
jjhelmus: Bazel (the build system tensorflow uses) seems to need these and I don't think the conda zlib will work.  I'll double check locally.

ci_support/run_docker_build.sh
ocefpaf: rerender? Then it is fine. (I saw the lines and assumed they were in `build.sh` but now I see where they are.)

jakirkham: Right that is how @pelson chose to implement it in PR ( https://github.com/conda-forge/conda-smithy/pull/134 ). I think it is the right call in the end as it makes the code human friendly.

recipe/yum_requirements.txt
ocefpaf: Don't hold the recipe b/c of this! I am just curious if `zlib` needs any changes to be fit for bazel.

recipe/build.sh
jakirkham: @183amir was trying to work on that in PR ( https://github.com/conda-forge/staged-recipes/pull/1019 ), but was having some issues. I don't recall what they were.

recipe/yum_requirements.txt
```diff
@@ -0,0 +1,5 @@
+java-1.8.0-openjdk-devel
+unzip
+zip
```
jakirkham: Was proposing we add `zip` and `unzip` based on a user request in PR ( https://github.com/conda-forge/docker-images/pull/20 ). Haven't gotten any feedback on it though. Should we be adding those?

recipe/yum_requirements.txt
jjhelmus: Yup, Bazel seems to ignore the conda installed zlib, so the yum zlib-devel is required.  There may be a method to adjust Bazel' search path but I had not been able to figure out how to do this.

ocefpaf: Since, right now, only `tensorflow` use `bazel` I would bother too much with this. If we start making heavy use of it though we should re-visit this.

jjhelmus:
Oops, I need to include the tar ball when building from source.  :laughing:

ocefpaf:
> Oops, I need to include the tar ball when building from source.  :laughing: 

What an awesome reviewer you got too. Noticed a bunch of small irrelevant details but missed that :smile_cat:

jjhelmus:
I am a bit worried about timing out CircleCI with this build. What is our limit there?

jakirkham:
~2hrs

jakirkham:
It looks like it might be building the examples. Do we have a way to disable that?

recipe/yum_requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+zlib-devel
+rsync
```
jakirkham: Just for my own interest, would be very curious to know what this is being used for.

jjhelmus: The build_pip_package script uses rsync to copy some files over into the whl file.  Without rsync the `bazel-bin/tensorflow/tools/pip_package/build_pip_package $SRC_DIR/tensorflow_pkg` command fails.

jjhelmus:
Not sure what went wrong with the build, will need to dig into this later.

jakirkham:
> Error: unexpected EOF from Bazel server.

Maybe a download issue. 😕

recipe/build.sh
183amir: I could not figure out what was wrong with bazel. If you want, please open a new pull in staged-recipes and iterate on bazel based on my recipe.

jjhelmus:
The Linux build worked :fireworks:.  Takes 52 minutes to build for a single Python version. Using a pre-build Bazel might cut 6-7 minutes off that but I even with that I do not think we can build all for all three Python version in a single run.  Is there a suggested method for splitting the build matrix into separate runs?

183amir:
Yes I saw @jakirkham do something similar in the Scipy feedstock

jakirkham:
Thanks for reminding me @183amir. Was meaning to write something earlier, but got caught up.

In the long run, I'd like us to leverage CircleCI parallelism to solve this problem. See this issue ( https://github.com/conda-forge/conda-smithy/issues/203 ) roughly outlining what that would look like. However, that is going to be a bit more work that I haven't had time to really dig into.

Short term solution would be more branches. That's what we did with SciPy. Basically add specialized skips for each so that each one builds a Python version. The maintenance will kind of suck, but at least it gets something working soon if you want to do that. See PR ( https://github.com/conda-forge/scipy-feedstock/pull/10 ) for a rough description of how this works.

jakirkham:
PS - Great job getting this build, @jjhelmus!

jakirkham:
Should add using `CPU_COUNT` is something worth trying too if `bazel` does parallel builds. Though it probably won't be enough. That would get you parallelism of 2 on Linux.

jjhelmus:
bazel [estimates the number of CPUs](https://www.bazel.io/docs/bazel-user-manual.html#flag--local_resources) (and RAM/IO capability) and we are explicitly using 2 jobs, so I do not think setting `CPU_COUNT` will help.

msarahan:
Inched this forward a tiny bit at https://github.com/jjhelmus/tensorflow-feedstock/pull/1 - just hacked it to use Conda's zlib library.

msarahan:
If anyone else tries this, you should be aware that running out of RAM does not crash.  Rather, it enters an infinite loom: https://github.com/tensorflow/tensorflow/issues/349

recipe/meta.yaml
```diff
@@ -2,8 +2,13 @@ package:
   name: tensorflow
   version: "0.9.0"
 
+source:
+  fn: v0.9.0.tar.gz
+  url: https://github.com/tensorflow/tensorflow/archive/v0.9.0.tar.gz
```
jakirkham: Kind of a small thing, but version templating might be a nice thing. Not a blocker by any means. Just mentioning it.

recipe/meta.yaml
```diff
@@ -2,8 +2,13 @@ package:
   name: tensorflow
   version: "0.9.0"
 
+source:
+  fn: v0.9.0.tar.gz
```
jakirkham: Maybe prefix with `tensorflow-*`. Kind of good to have a unique name for local source caches. Again not a blocker. Just a suggestion.

recipe/yum_requirements.txt
jakirkham: So we have `zip` and `unzip` as conda-forge packages now.

ivoflipse:
Tensorflow should now also be able to build from source for Windows: https://github.com/tensorflow/tensorflow/issues/17

The downside of course is that it will probably take too long

jjhelmus:
Succeeded by #74

ocefpaf:
Note that I removed a lot of the macOS and Windows stuff from this PR that we may still use in the future. #74 only preserved the Linux part of the source build.

PR #14, opened 2016-09-01
Update to Tensorflow 0.10.0rc0

tcwalther:
When you go on www.tensorflow.org and click on Get Started, Google now recommends you to download Tensorflow 0.10.0rc0 by default. Even though it's still an RC, it seems to be the official package now, so I think it would be good to add this to conda-forge.

This is my first PR on conda-forge. I'm not quite sure how to test recipes for conda package management before making a PR - it seems like [making the PR will trigger the test](http://conda-forge.github.io/#update_recipe). If there is a way to test this before creating the PR, I'd greatly appreciate a hint on how to do it.


tcwalther:
I figured out that running `./ci_support/run_docker_build.sh`on a machine with Docker installed does the trick. That's actually fairly convenient :-).

jjhelmus:
With tensorflow `0.9.0` we choose not to package the release candidate, see #8, but `0.10.0rc0` has been out for over a month with no indication of a non-rc release coming.  Given this I am in favor of packaging the `rc` releases.  Do other agree?

jjhelmus:
Should mention that conda seems to order `0.x.0` releases ahead of `0.x.0rc0` so `conda install -c conda-forge tensorflow` will install `0.10.0` and not `0.10.0rc0` once it has been released.

jjhelmus:
@tcwalther It looks like the build matrices where removed from the CI files so only a single Python version is being built.  Can you rerender the recipe with conda smithy.  I believe that should fix this issue.

tcwalther:
I reran it, I hope this fixed it.

jjhelmus:
The Linux build on Circle CI seems to be in order but there still is only a single OS X build on TravisCI.  Did you happen to forget to add the `.travis.yml` file to the commit after running `conda smithy regenerate`?

aaronzs:
@tcwalther tensorflow 0.10.0 is released.

jjhelmus:
closing now that version 0.10.0 has been released and is available through `conda-forge`.  See #16

PR #20, opened 2016-11-01
Tensorflow 0.11.0rc2 recipe

jjhelmus:


jjhelmus:
Although these are release candidate versions of tensorflow if you install the package using the instructions in the documentation these are the versions that would be installed. I think we should try to package them, although keeping up with the pace of releases is proving to be challenging.

PR #24, opened 2016-11-30
WIP: Build Linux from source using cmake

jjhelmus:
Build tensorflow from source using cmake on Linux.  Limiting the build to Python 3.5 to limit CI needs.  Removed Windows and OS X build for the time being.


jjhelmus:
Oops, probably should use 8 jobs during the build!  Was using that for testing.  WIll wait for the current test to finish and then push up the change.

jjhelmus:
Unfortunately using `cmake` to build `tensorflow` only seems to work for Python 3.5.  But at least it is a start.

jjhelmus:
Closing.  Building from source would be nice but this need large updates before it works in general.

Issue #35, opened 2017-06-26
Release 1.2


hajapy:
Tensorflow 1.2 has been released: https://github.com/tensorflow/tensorflow/releases

Looking through the requirements for the [pypi wheels](https://pypi.python.org/pypi/tensorflow), I noticed a possible issue. This release brings in bleach 1.5.0 and html5lib (==0.9999999) and on conda-forge we are missing a build for this exact release of html5lib. So it looks like we need to build an old version on the html5lib feedstock if we want working run requirements for the tensorflow 1.2 recipe. 

As far as I can tell, this bleach / html5lib requirement only serves one feature of tensorboard: https://github.com/tensorflow/tensorboard/blob/f024dc5078f68cb926a4c341eeabc5679afa29cb/tensorboard/plugins/text/text_plugin.py

adrianchia:
I think backports.weakref is also missing

hajapy:
You're right, thanks for opening the PR at staged-recipes. Should help us get to 1.2 sooner :)

PR #36, opened 2017-06-27
Update tensorflow to 1.2.0

adrianchia:
Fixes https://github.com/conda-forge/tensorflow-feedstock/issues/35


hajapy:
A few things:

1. https://github.com/tensorflow/tensorflow/issues/11063#issuecomment-311196587 means we can probably use markdown >= 2.6.8
2. The bleach requirement is fine, but you should also add the explicit html5lib ==0.9999999 requirement, otherwise it will use a newer version that may not work
3. Why was windows disabled completely? We should be able to get pip wheels for py35 and py36 now: https://pypi.python.org/pypi/tensorflow
4. Why are there no longer 3 cases to embark on for mac/linux? 
5. This has carried forward for a while now, but the build requirements could be pared down substantially since we just use python, pip and (maybe) wheel to repackage the binaries.

adrianchia:
I believe conda smity rerender caused the removal of 3 environments, let me revert this to previous version along with other changes.

recipe/meta.yaml
```diff
@@ -26,13 +26,21 @@ requirements:
     - six >=1.10.0
     - wheel
     - werkzeug >=0.11.10
+    - markdown >=2.6.8
+    - bleach ==1.5.0
+    - backports.weakref ==1.0rc1
+    - html5lib ==0.9999999
```
jakirkham: Admittedly much of this is from before your PR, but could you please drop everything from `build` that isn't strictly required for building the package. Probably can get away with `python`, `pip` only. Then please try re-rendering. Also please make sure you have the latest `conda-smithy` and `conda-build-all`.

jakirkham: note: This is already done.

adrianchia:
Recipe updated to include Python 3.6 Windows build.

hajapy:
Looks good to me, other changes (small test script & checksum validation) can come in a separate PR

hajapy:
1.2.1 has been released now with the fix to the markdown requirement: https://github.com/tensorflow/tensorflow/releases/tag/v1.2.1

jakirkham:
Went ahead and merged this. Hope that is ok @jjhelmus. Feel free to send an update PR @hajapy.

PR #45, opened 2017-11-07
Update to 1.4.0

dhirschfeld:
Is this all that's required?

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: tensorflow
-  version: "1.3.0"
+  version: "1.4.0"
 
 build:
   number: 0
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -22,15 +22,17 @@ requirements:
     - pip
   run:
     - python
+    - enum34 >=1.1.6 # [py<34]
     - mock >=2.0.0
-    - numpy >=1.11.0
-    - protobuf >=3.2.0
+    - numpy >=1.12.1
+    - protobuf >=3.3.0
     - six >=1.10.0
     - werkzeug >=0.11.10
     - markdown >=2.6.8
     - bleach ==1.5.0
     - backports.weakref ==1.0rc1
     - html5lib ==0.9999999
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Looks like the Travis py27 build failed with:
```
ImportError: No module named enum
```

hajapy:
It seems this is now required in 1.4 for py<34

https://github.com/tensorflow/tensorflow/blob/v1.4.0/tensorflow/tools/pip_package/setup.py#L35

It is used here: 
https://github.com/tensorflow/tensorflow/blob/v1.4.0/tensorflow/python/ops/variable_scope.py#L24

Since the import test is skipped for linux and windows py27 is not built, only travis (os x) shows the error but it would be an issue with for linux as well.

hajapy:
I would also add that a number of the required run time dependencies are outdated as compared to the setup.py linked above (a few are related to tensorboard which is not bundled in the wheel anymore). 

When https://github.com/conda-forge/staged-recipes/pull/3866 is done we can update the run requirements to include tensorboard.

jerowe:
I didn't even see this and naively opened a PR for 1.4. ;-)

How is the progress on this? The newest tensorflow/models repo says it needs 1.4, and I would like to upgrade my env.

hajapy:
@dhirschfeld think you will come back to wrapping this up?

dhirschfeld:
@hajapy - I gave it a go, we'll see what the CI thinks...

hajapy:
Awesome, thanks! Looks like CI is happy. I still think we should remove all the tensorboard only run dependencies as those are already part of the tensorboard recipe. After that change, will be ready for merge!

dhirschfeld:
I won't have time to get to this until Friday now so feel free to edit & push to my branch if you want...

hajapy:
@dhirschfeld https://github.com/dhirschfeld/tensorflow-feedstock/pull/1

hajapy:
ping @conda-forge/core could someone please merge this?

danielfrg:
Thanks for the update!

PR #47, opened 2017-12-10
Remove wrong tensorboard entry point

sdvillal:
This is causing the tensorboard command to be broken.
See: https://github.com/conda-forge/tensorboard-feedstock/issues/1


sdvillal:
I also bumped to version 1.4.1

sdvillal:
I'm a bit lost on how to fix the tensorboard CLI to work on windows. How should we modify the [entry point back in the tensorboard recipe](https://github.com/loopbio/tensorboard-feedstock/blob/0d0914355df2af40ee659da09ac0a12ed5bfc573/recipe/meta.yaml#L16)? Just remove it for windows?

It is just impossible for me to test without pushing new tensorboard builds to conda-forge main channel, which I can do but I also think would be preposterous. I really think we should make it so that we have a sane dependency tree (tensorboard depends on tensorflow and not the other way around), regardless of the mistaken inverse dependency in pypi.

sdvillal:
Another possibility is to remove conda-forge tensorboard package, and make our tensorflow package contain tensorboard (i.e. pip install tensorboard in this recipe too). Even if google decided to split these, they did it the wrong way. We should be running in google clusters, not in google clusterfucks.

hajapy:
May be some clue as to how to fix by looking at the setup.py files for both tensorflow/tensorboard upstream:

https://github.com/tensorflow/tensorflow/blob/v1.4.1/tensorflow/tools/pip_package/setup.py#L69
https://github.com/tensorflow/tensorboard/blob/0.4.0-rc3/tensorboard/pip_package/setup.py#L44

In tensorboard the entry point is `tensorboard = tensorboard.main:run_main`
In tensorflow the entry point is `tensorboard = tensorboard.main:main`

Also, note we are not properly exposing the `saved_model_cli = tensorflow.python.tools.saved_model_cli:main` entrypoint in tensorflow recipe. 

Lastly, it should be possible for you to set up your AppVeyor project for you fork of this feedstock to allow Remote Desktop to the build machine via RDP. In that case you could iterate on the recipe until you get it working in windows without need the setup a windows machine/environment yourself.

jakirkham:
@conda-forge-admin, please rerender.

sdvillal:
This got old and very wrong. Hopefully it will be tackled in #54.

Issue #11, opened 2016-08-24
GLIBC 2.17 not found


rainwoodman:
I am quite new to conda-forge. It looks like a problem on my end -- there must be a very simple fix to this? Could someone give some light? Thanks.

```
bash-4.1$ python                                                                                                                      
impPython 3.5.1 |Anaconda custom (64-bit)| (default, Dec  7 2015, 11:16:01)                                                           
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux                                                                                       
Type "help", "copyright", "credits" or "license" for more information.                                                                
>>> import tensorflow                                                                                                                 
Traceback (most recent call last):                                                                                                    
  File "<stdin>", line 1, in <module>                                                                                                 
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/site-packages/tensorflow/__init__.py", line 23, in <module>    
    from tensorflow.python import *                                                                                                   
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/site-packages/tensorflow/python/__init__.py", line 48, in <modu
le>                                                                                                                                   
    from tensorflow.python import pywrap_tensorflow                                                                                   
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/site-packages/tensorflow/python/pywrap_tensorflow.py", line 28,
 in <module>                                                                                                                          
    _pywrap_tensorflow = swig_import_helper()                                                                                         
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/site-packages/tensorflow/python/pywrap_tensorflow.py", line 24,
 in swig_import_helper                                                                                                                
    _mod = imp.load_module('_pywrap_tensorflow', fp, pathname, description)                                                           
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/imp.py", line 242, in load_module                              
    return load_dynamic(name, filename, file)                                                                                         
  File "/project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python3.5/imp.py", line 342, in load_dynamic                             
    return _load(spec)                                                                                                                
ImportError: /lib64/libc.so.6: version `GLIBC_2.17' not found (required by /project/projectdirs/m779/yfeng1/envs/tensorflow/lib/python
3.5/site-packages/tensorflow/python/_pywrap_tensorflow.so) 
```

[...]

PR #53, opened 2018-05-08
Updates version and dependencies of Tensorflow to 1.8.0

gilbertfrancois:


recipe/meta.yaml
```diff
@@ -28,8 +28,12 @@ requirements:
     - numpy >=1.12.1
     - protobuf >=3.4.0
     - six >=1.10.0
-    - tensorboard >=1.5.0,<1.6.0
     - backports.weakref >=1.0rc1
+    - astor
+    - grpcio
+    - termcolor
+    - gast
+    - tensorboard >=1.7.0,<1.8.0
```
dhirschfeld: Is this valid? `tensorflow` clearly pins `tensorboard` to `>= 1.8.0, < 1.9.0` so you'd assume there's a good reason for that?
https://github.com/tensorflow/tensorflow/blob/v1.8.0/tensorflow/tools/pip_package/setup.py#L34-L43

I'd be inclined to respect the pinning as per their `setup.py` and just update `tensorboard`

recipe/meta.yaml
```diff
@@ -28,8 +28,12 @@ requirements:
     - numpy >=1.12.1
     - protobuf >=3.4.0
     - six >=1.10.0
-    - tensorboard >=1.5.0,<1.6.0
     - backports.weakref >=1.0rc1
+    - astor
+    - grpcio
+    - termcolor
+    - gast
```
jjhelmus: These new requirements should all have lower version bounds as specified in the projects setup.py specifically:
```
astor >=0.6.0
grpcio >=1.8.6
termcolor >=1.1.0
gast >=0.2.0

File: recipe/meta.yaml [hunk 3]
```diff
@@ -28,8 +28,12 @@ requirements:
     - numpy >=1.12.1
     - protobuf >=3.4.0
     - six >=1.10.0
-    - tensorboard >=1.5.0,<1.6.0
     - backports.weakref >=1.0rc1
```
jjhelmus: This is only required on python 2.  A selector should be added:
```
    - backports.weakref >=1.0rc1  # [py2k]
```

recipe/meta.yaml
```diff
@@ -28,8 +28,12 @@ requirements:
     - numpy >=1.12.1
```
jjhelmus: For the 1.8.0 release this should likely be updated to >=1.13.3

dhirschfeld:
***xref:*** https://github.com/conda-forge/tensorboard-feedstock/pull/7

gilbertfrancois:
Thank you so much for the valuable comments. They were very helpful. I've updated the requirements in line with the `setup.py` as recommended. The current automated build fails because it cannot find tensor board 1.8.0 yet. It should work as soon as the tensorboard PR is accepted.

recipe/meta.yaml
```diff
@@ -23,13 +23,17 @@ requirements:
...
     - six >=1.10.0
-    - tensorboard >=1.5.0,<1.6.0
-    - backports.weakref >=1.0rc1
+    - protobuf >=3.4.0
+    - tensorboard >=1.8.0, <1.9.0
+    - termcolor >=1.1.0
+    - grpcio >=1.8.6
+    - mock >=2.0.0                # [py27]
+    - enum34 >=1.1.6              # [py<34]
+    - backports.weakref >=1.0rc1  # [py<34]
```
dhirschfeld: Should be `- backports.weakref  # [py2k]` I think

recipe/meta.yaml
```diff
@@ -23,13 +23,17 @@ requirements:
...
+    - gast >=0.2.0
+    - numpy >=1.13.3
     - six >=1.10.0
-    - tensorboard >=1.5.0,<1.6.0
-    - backports.weakref >=1.0rc1
+    - protobuf >=3.4.0
+    - tensorboard >=1.8.0, <1.9.0
+    - termcolor >=1.1.0
+    - grpcio >=1.8.6
+    - mock >=2.0.0                # [py27]
```
dhirschfeld: IIUC `mock` is only required for testing so can be removed from the run requirements.

recipe/meta.yaml
gilbertfrancois: In the setup.py it is coded as:
```
# weakref.finalize and enum were introduced in Python 3.4
if sys.version_info < (3, 4):
  REQUIRED_PACKAGES.append('backports.weakref >= 1.0rc1')
  REQUIRED_PACKAGES.append('enum34 >= 1.1.6')
```
It is still better to use `# [py2k]`?

recipe/meta.yaml
gilbertfrancois: Thanks, I'll do that.

gilbertfrancois:
I keep having issues starting Tensorboard from the cli, even after this change: [0e2fe37](https://github.com/conda-forge/tensorflow-feedstock/pull/53/commits/0e2fe37f9831685bba20ae9d784108915215a1ad). Does anyone have an idea how to solve it?

recipe/meta.yaml
dougalsutherland: They have exactly the same effect on conda-forge, doesn't matter.

gilbertfrancois:
The build for macOS python 2.7 gave an error. Do I have to do `conda smithy rerender` again? Or does this error have another cause?

```The command "upload_or_check_non_existence ./recipe conda-forge --channel=main -m ./.ci_support/${CONFIG}.yaml" exited with 0.```

**Comment by gilbertfrancois** [comment]
I removed `mock` from the `requirements: run:` section but forgot to put it back in the `test: requires:` section for testing with py27.  (Thank you @dhirschfeld for the hint!). Hopefully all builds will be green now :)

**Comment by dhirschfeld** [comment]
Maybe intermittent. I'll close & reopen to trigger the CI again...

**Comment by dhirschfeld** [comment]
LGTM & tests pass so happy to merge this. I'll hold off for now to give others time to review/comment...

File: recipe/meta.yaml [hunk 7]
```diff
@@ -14,7 +14,16 @@ build:
   skip: True  # [py26]
   skip: True  # [py33]
   entry_points:
-    - tensorboard = tensorflow.tensorboard.tensorboard:main
+    - freeze_graph = tensorflow.python.tools.freeze_graph:run_main
+    - toco_from_protos = tensorflow.contrib.lite.toco.python.toco_from_protos:main
+    - toco = tensorflow.contrib.lite.toco.python.toco_wrapper:main
+    - saved_model_cli = tensorflow.python.tools.saved_model_cli:main
+    # We need to keep the TensorBoard command, even though the console script
+    # is now declared by the tensorboard pip package. If we remove the
```
sdvillal: Isn't conda now the one responsible for creating correctly the entrypoints from now on?

sdvillal:
I would add entrypoint tests, as commented in https://github.com/conda-forge/tensorflow-feedstock/pull/54#issuecomment-387942011.

Also I would squash to a single commit before merging.

gilbertfrancois:
Added tests for the cli applications, as suggested by @sdvillal.

dhirschfeld:
The error seems a bit concerning:
```
Executing transaction: ...working... done
+ tensorboard --help
+ grep USAGE
  File "/home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/site-packages/tensorflow/python/pywrap_tensorflow.py", line 58, in <module>
    from tensorflow.python.pywrap_tensorflow_internal import *
  File "/home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 28, in <module>
    _pywrap_tensorflow_internal = swig_import_helper()
  File "/home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 24, in swig_import_helper
    _mod = imp.load_module('_pywrap_tensorflow_internal', fp, pathname, description)
  File "/home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/imp.py", line 243, in load_module
    return load_dynamic(name, filename, file)
  File "/home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/imp.py", line 343, in load_dynamic
    return _load(spec)
ImportError: /lib64/libc.so.6: version `GLIBC_2.16' not found (required by /home/conda/feedstock_root/build_artifacts/tensorflow_1525944978319/_test_env_placeho/lib/python3.5/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so)
```

recipe/meta.yaml
```diff
@@ -50,6 +50,11 @@ test:
     # Skip the import test on Linux as wheel file require a more recent
     # version of GLIBC++ than the VM used to build and test package.
     - tensorflow  # [not linux]
+  commands:
+    - tensorboard --help |grep USAGE         # [not win]
+    - freeze_graph --help |grep usage        # [not win]
+    - toco_from_protos --help |grep usage    # [not win]
+    - saved_model_cli --help |grep usage     # [not win]
```
dhirschfeld: I'm not sure why `| grep usage` is required. That's explicitly testing for something we don't care about (the word usage in the help text) and seems redundant.

If the entrypoint isn't there the command will fail and hence the test will fail. If the entrypoint is there but the word *usage* isn't in the text, the test will also fail which isn't something we want. Maybe that would never happen in practice but it's still a redundant test and is preventing it being tested on Windows.

I'd be inclined to remove the `|grep usage` and test on all supported platforms.

recipe/meta.yaml
dhirschfeld: I think this comment isn't particularly valid for the conda packages so can be removed. Not a blocker though.

gilbertfrancois:
Noted. I change it later today.

dhirschfeld:
#11 seems related

sdvillal:
End of road for entrypoint testing while we keep just repackaging the tensorflow wheels. From 1.5 tensorflow wheels are built against newer glibc - so they cannot work in our centos image. So I would revert the testing changes. Sorry to see we cannot automate testing.

As a side note, ```tensorboard --help``` alone errcodes to 1 and would make our test fail. That is why you need something like the grep pipe.

dougalsutherland:
Yeah, that's the problem with us not building from source. :\

Would be good to run the tests on Mac and Windows, though.

gilbertfrancois:
I did not see the above mentioned error on Linux (ubuntu 16.04.3 LTS) and macOS 10.13 when I tested on my local machines. Thank you @sdvillal for addressing the cause. I restricted now the test to macOS, since I have no possibilities to test it on Windows.

dhirschfeld:
@gilbertfrancois - please also test on Windows. Just let the CI test for you - that's what it's there for. If it fails on Windows it's telling us something important, something we need to fix.

gilbertfrancois:
I added the tests for Windows. 
- `tensorboard` passed with a pipe to ` findstr`
- `freeze_graph` passed
- `toco_from_protos` failed
- `saved_model_cli` unknown (did not reach this test in CI yet)

```
(C:\bld\tensorflow_1525984894103\_test_env) C:\bld\tensorflow_1525984894103\test_tmp>toco_from_protos --help 
565Traceback (most recent call last):
566  File "C:\bld\tensorflow_1525984894103\_test_env\Scripts\toco_from_protos-script.py", line 6, in <module>
567    from tensorflow.contrib.lite.toco.python.toco_from_protos import main
568ImportError: No module named 'tensorflow.contrib.lite.toco.python'
569
570(C:\bld\tensorflow_1525984894103\_test_env) C:\bld\tensorflow_1525984894103\test_tmp>IF 1 NEQ 0 exit 1 
571
```

I'm kind of stuck here, since I've no Windows machine and cannot debug. Can it be that `contrib.toco` is not included in the windows package?

sdvillal:
It seems no toco for win, so I would just remove that test and be happy. Everything else looks great.

gilbertfrancois:
Great! Thanks for checking.

dhirschfeld:
I'm happy with it too and think it's ready for merging. Will do so in a few hours if no one beats me to it...

dhirschfeld:
> *As a side note, `tensorboard --help` alone errcodes to 1 and would make our test fail. That is why you need something like the grep pipe.*

Tested and confirmed on Windows. That doesn't seem like best practice :|

dougalsutherland:
Since `toco` / `toco_from_proto` don't work on Windows, let's just not include the entry points, right?

Also added a test of the `toco` CLI.

gilbertfrancois:
It seems like toco is not useful as cli application at all. I saw this on macOS:
```
$ toco --help
TOCO from pip install is currently not working on command line.
Please use the python TOCO API or use
bazel run tensorflow/contrib/lite:toco -- <args> from a TensorFlow source dir.
```
I removed the entrypoint completely now. No clue why it is in the `setup.py` in the first place.

dougalsutherland:
Merging. Thanks @gilbertfrancois / @dhirschfeld / @sdvillal / @jjhelmus!

dhirschfeld:
Thanks @gilbertfrancois for the PR & your perseverance getting it right!

gilbertfrancois:
Big thanks to @dhirschfeld, @dougalsutherland, @sdvillal, @jjhelmus for the great comments and guidance. It was a really nice welcome in the world of _conda-forge_. Thumbs up.

romrozen:
Hey guy, so if I understand correctly toco is not available on Windows yet?
So if I want to transform TF model to TFLite, the options are Mac and Linux?

Issue #48, opened 2018-01-04
No Module named tensorflow.tensorboard


AustinT:
I installed tensorflow/tensorboard yesterday using:

`conda install -c conda-forge tensorflow`

Tensorflow works fine but tensorboard does not work. Here is my code from the command line:

```
C:\Users\Austin>tensorboard --logdir="C:\Users\Austin\Desktop\handson-ml\tf_logs"  
Traceback (most recent call last):  
  File "C:\ProgramData\Anaconda3\Scripts\tensorboard-script.py", line 3, in <module>  
    import tensorflow.tensorboard.tensorboard  
ModuleNotFoundError: No module named 'tensorflow.tensorboard'
```

I believe the issue is that the line `import tensorflow.tensorboard.tensorboard` in _tensorboard-script.py_ needs to be edited because it refers to the old tensorflow structure where tensorboard was not its own package, but I am not sure exactly what it should be changed to. Any help would be greatly appreciated.

## System information
- Windows 10
- Python 3.6 (via Anaconda)
- Tensorflow version 1.4.0 (default latest)

*Note:* I also had this issue on my other PC, but uninstalling it and installing via pip as outlined on the tensorflow website worked. (https://www.tensorflow.org/install/install_windows)

gilbertfrancois:
Please update to version 1.8.0. The issue has been resolved.

dougalsutherland:
(Note that the builds for 1.8 aren't quite online yet, but should be in the next few hours.)

PR #52, opened 2018-04-23
Release 1.6.0

jakirkham:

PR #54, opened 2018-05-09
update to 1.8.0

dougalsutherland:
Based on @gilbertfrancois's changes in #53, but avoiding merge conflicts with #52.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: tensorflow
-  version: "1.6.0"
+  version: "1.8.0"
 
 build:
   number: 0
@@ -14,7 +14,7 @@ build:
   skip: True  # [py26]
   skip: True  # [py33]
   entry_points:
-    - tensorboard = tensorflow.tensorboard.tensorboard:main
+    - tensorboard = tensorboard.main:run_main
 
 requirements:
   build:
@@ -31,7 +31,7 @@ requirements:
     - numpy >=1.13.3
     - protobuf >=3.4.0
     - six >=1.10.0
-    - tensorboard >=1.6.0,<1.7.0
+    - tensorboard >=1.8.0,<1.9.0
     - termcolor >=1.1.0
     - backports.weakref >=1.0rc1  # [py2k]
 
@@ -54,3 +54,5 @@ extra:
     - ghego
     - hajapy
     - dhirschfeld
+    - gilbertfrancois
+    - dougalsutherland


dougalsutherland:
@gilbertfrancois: Can you clarify what you meant by https://github.com/conda-forge/tensorflow-feedstock/pull/53#issuecomment-387744099? I haven't tested locally, but does this work now that the updated `tensorboard` package is available?

gilbertfrancois:
The packages and dependencies of #53 are all ok. However, when starting `$ tensorboard --logdir [dir]` it throws an error. The version `1.6.0` is broken too, see below:

```
(test) ubuntu@ip-172-31-8-108:~$ conda list tensor*
# packages in environment at /home/ubuntu/anaconda3/envs/test:
#
# Name                    Version                   Build  Channel
tensorboard               1.6.0                    py36_0    conda-forge
tensorflow                1.6.0                    py36_0    conda-forge

(test) ubuntu@ip-172-31-8-108:~$ tensorboard --logdir /data/logs/tfscratch/ 
  File "/home/ubuntu/anaconda3/envs/test/bin/tensorboard", line 7, in <module> 
    from tensorflow.tensorboard.tensorboard import main 
ModuleNotFoundError: No module named 'tensorflow.tensorboard' 
```
I think I've found the solution and will update both tensorflow and tensorboard `meta.yaml` files. It is caused by the split they made in the packages but still have cross dependencies. I test before I push the new code.

gilbertfrancois:
@dougalsutherland: The Tensorboard crash has been solved. Both the tensorflow and tensorboard `meta.yaml` files have small changes. Please add this:
```
  entry_points:
    - freeze_graph = tensorflow.python.tools.freeze_graph:run_main
    - toco_from_protos = tensorflow.contrib.lite.toco.python.toco_from_protos:main
    - toco = tensorflow.contrib.lite.toco.python.toco_wrapper:main
    - saved_model_cli = tensorflow.python.tools.saved_model_cli:main
    # We need to keep the TensorBoard command, even though the console script
    # is now declared by the tensorboard pip package. If we remove the
    # TensorBoard command, pip will inappropriately remove it during install,
    # even though the command is not removed, just moved to a different wheel.
    - tensorboard = tensorboard.main:run_main
```
to the build section in `meta.yaml`. I don't know (yet) how to add it in your pull request. I did it in #53 as well, but that build seems to have a problem with building the package for macOS py27 on Travis.

Thank you.

sdvillal:
I agree with @gilbertfrancois regarding the entrypoints. I think we should start adding here tests for them, in particular for tensorboard.

At least for linux and mac it should be something like:

```yaml
commands:
    - tensorboard --help | grep USAGE
```

See https://github.com/conda-forge/tensorboard-feedstock/pull/8#issuecomment-387940850

I think this should work now without much hassle, and then we would close #48 and #47.

gilbertfrancois:
Please consider reviewing and merging #53.

sdvillal:
Commented the same there. I have no permisions to close/merge in this repo, but I think we should close this PR to avoid duplicating efforts.

PR #55, opened 2018-05-12
Require mock

jschueller:
Seems tensorflow requires mock at runtime, not just testing anymore:
https://github.com/tensorflow/tensorflow/search?utf8=%E2%9C%93&q=%22import+mock%22&type=Code
Encoutered this while creating gpflow recipe https://github.com/conda-forge/staged-recipes/pull/5847

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: "1.8.0"
 
 build:
-  number: 0
+  number: 1
   #  Google supplies whl files on PyPI for:
   # - Linux: Python 2.7, 3.3, 3.4, 3.5 and 3.6
   # - OS X: Python 2.7, 3.5 and 3.6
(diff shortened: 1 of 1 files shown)


dougalsutherland:
So it seems like that's only used in some test code, which I don't see imported anywhere. I also don't see a failed build on the gpflow recipe. What actually failed?

(Not like it's a big dependency, though, so maybe we should just include it anyway to be safe.)

jschueller:
The build doesnt fail as we require mock for testing, but 'import tensorflow' does import mock at some point as the test module is imported by the root init.py

dhirschfeld:
Sorry, my fault - I didn't search hard enough for indirect imports. I think that's actually *terrible* practice, but it is what it is.

I also surprised that the `import tensorflow` test didnt' pick that up as it seems to be in the root [`__init__.py`](https://github.com/tensorflow/tensorflow/blob/703f47614cf9ca6af9bc9fc71db1880b1fbe3f3c/tensorflow/python/__init__.py#L121)

dhirschfeld:
The reason the test may have passed is that `py27` isn't tested on Windows:
https://github.com/conda-forge/tensorflow-feedstock/blob/9c3793b408745b53499a30d458025162c33a154d/recipe/meta.yaml#L12

...and the test is actually skipped on linux:
https://github.com/conda-forge/tensorflow-feedstock/blob/9c3793b408745b53499a30d458025162c33a154d/recipe/meta.yaml#L44-L47

I would've thought OSX might have picked it up though 😕

dhirschfeld:
In any case, I'm happy to merge this. Will let others review before doing so though...

PR #58, opened 2018-07-17
Upgrade to Tensorflow v1.9.0

FarhanTejani:
Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

We will have to wait for the [tensorboard upgrade](https://github.com/conda-forge/tensorboard-feedstock/pull/9) to merge before the tests will pass.


FarhanTejani:
Seems like the tensorboard test is failing? Although I'm not sure why...

recipe/meta.yaml
```diff
@@ -32,11 +32,11 @@ requirements:
     - numpy >=1.13.3
     - six >=1.10.0
     - protobuf >=3.4.0
-    - tensorboard >=1.8.0, <1.9.0
+    - tensorboard 1.9
```
dhirschfeld: Perhaps that should that be `1.9*`? i.e. any patch release in the `1.9` version.

dougalsutherland: `1.9` also matches `1.9.*`. It's better than `>=1.9.0,<1.10` though in that the latter would also match `1.10.0rc1 ` or similar.

dougalsutherland:
Merging without the changes I attempted in #59 because I couldn't quite get them working.

PR #60, opened 2018-07-23
Update meta.yaml

LarsDu:
I found that Tensorflow  requires a newer version of glibc than is available at my institution's old CentOS computing cluster. Adding libstdcxx-ng seems to fix this problem.

Checklist
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: "1.9.0"
 
 build:
-  number: 0
+  number: 1
   #  Google supplies whl files on PyPI for:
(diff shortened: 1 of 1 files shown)


jjhelmus:
This may fix issues with newer `GLIBCXX` symbols but it will not fix any issue with `GLIBC` symbols as these are provided by your system libc.  The tensorflow package in the `defaults` channel will  work on CentOS 6 if that is of interest.

PR #62, opened 2018-08-14
Update tensorflow version to 1.10.0

kyleabeauchamp:
Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy`

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -49,12 +49,12 @@ test:
     # https://github.com/conda-forge/tensorflow-feedstock/pull/62
      - tensorflow  # [win]
   commands:
-    # - tensorboard --help | grep USAGE        # [osx]
-    - tensorboard --help | findstr USAGE     # [win]
-    - freeze_graph --help                    # [win]
+    # - tensorboard --help | grep USAGE       # [osx]
+    - tensorboard --help                      # [win]
+    - freeze_graph --help                     # [win]
     #- toco_from_protos --help                # [osx]
     #- toco --help                            # [osx]
-    - saved_model_cli --help                 # [win]
+    - saved_model_cli --help                  # [win]
 
 about:
   home: http://tensorflow.org/
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -42,19 +42,19 @@ requirements:
 
 test:
 
-  # imports:
+  imports:
     # Skip the import test on Linux as wheel file require a more recent
     # version of GLIBC++ than the VM used to build and test package.
     # Also skip on OSX due to OSX version requirements, see
     # https://github.com/conda-forge/tensorflow-feedstock/pull/62
-    # - tensorflow  # [not linux]
+     - tensorflow  # [win]
   commands:
-    - tensorboard --help | grep USAGE        # [osx]
+    # - tensorboard --help | grep USAGE        # [osx]
     - tensorboard --help | findstr USAGE     # [win]
-    - freeze_graph --help                    # [not linux]
-    - toco_from_protos --help                # [osx]
-    - toco --help                            # [osx]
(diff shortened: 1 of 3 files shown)


recipe/meta.yaml
```diff
@@ -33,7 +33,7 @@ requirements:
     - numpy >=1.13.3
     - six >=1.10.0
     - protobuf >=3.4.0
```
jjhelmus: The protobuf requirement has increased to >=3.6.0

kyleabeauchamp:
Shoot, looks like we need to fix and merge:

https://github.com/conda-forge/libprotobuf-feedstock/pull/10

then 

https://github.com/conda-forge/protobuf-feedstock/pull/52/files

kyleabeauchamp:
It's also unclear to me what the difference is between protobuf and libprotobuf.  They both are installing from the same tarball...

jjhelmus:
I'm working on the update to protobuf.  Should have something available by the end of the day.  

The `libprotobuf` package is the C++ library, `protobuf` is the Python module.  They share the same source code but have different builds.

kyleabeauchamp:
Makes sense.  Thanks!

jjhelmus:
Update to libprotobuf, conda-forge/libprotobuf-feedstock#11

jjhelmus:
protobuf 3.6.0 is now available. I'm closing and restarting this PR to trigger the CIs.

dougalsutherland:
Travis is saying `Failed to load the native TensorFlow runtime.` on `import tensorflow`. Are Google's builds now only supporting newer OS versions than we're using there too?

jjhelmus:
It looks like the Google wheels require Mac OS X 10.12.  From the Travis log:
```
ImportError: dlopen(/Users/travis/miniconda3/conda-bld/tensorflow_1534344511854/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/python2.7/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so, 6): Symbol not found: _SecKeyCopyExternalRepresentation
  Referenced from: /Users/travis/miniconda3/conda-bld/tensorflow_1534344511854/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/python2.7/site-packages/tensorflow/python/_pywrap_tensorflow_internal.so (which was built for Mac OS X 10.12)
```

jjhelmus:
The official Tensorflow [mac install page](https://www.tensorflow.org/install/install_mac) indicates macOS 10.12.6 (Sierra) or higher is required.

dougalsutherland:
Okay, so we should remove `import tensorflow` from the tests on Mac too, and maybe slightly more strongly consider building it ourselves. :|

dougalsutherland:
@kyleabeauchamp Let's still import on Windows, no?

kyleabeauchamp:
Anyone have any clue which line of the windows test is failing?  The output is not really identifying the failure to me...

dougalsutherland:
Looks like `tensorboard --help` didn't contain the string `USAGE`. Could be for a couple of reasons; maybe try commenting out the `findstr` bit to see the output?

kyleabeauchamp:
Build is green.  Are you ok with just checking that the tensorboard help runs, with no grep?

dougalsutherland:
In older versions, `tensorboard --help` returned 1 (which is why the grep was in there). It looks like that's not the case anymore (thankfully), so this is good.

PR #74, opened 2019-08-06
Build all

ocefpaf:
This is a test to see if we can build Linux from source and wheels for the rest.

If this works it is an alternative for #69.


recipe/meta.yaml
```diff
@@ -72,14 +83,14 @@ requirements:
     - gast >=0.2.0
     - keras-applications >=1.0.6
     - keras-preprocessing >=1.0.5
-    - {{ pin_compatible('numpy') }}
+    - {{ pin_compatible('numpy') }}  # [linux]
```
jjhelmus: If numpy is behind a selector, the dependency on other platforms needs to be listed explicitly:
```
- numpy >=1.13.3   # [not linux]
```

ocefpaf: Thanks! Missed that one.

recipe/build.sh
```diff
@@ -1,23 +1,67 @@
...
+    mkdir -p $SRC_DIR/tensorflow_pkg
+    bazel-bin/tensorflow/tools/pip_package/build_pip_package $SRC_DIR/tensorflow_pkg
+
+    # install the whl using pip
+    pip install --no-deps $SRC_DIR/tensorflow_pkg/*.whl
+
+    # The tensorboard package has the proper entrypoint
+    rm -f ${PREFIX}/bin/tensorboard
+else
+  $PYTHON -m pip install --no-deps -vv $fn
```
ocefpaf: This does not work :-(
Looks like the variable for the name is not available here.
Also, if I try to use a `script` section in the recipe conda-build complains that I have a build script and a script line. Not sure what is the best way to fix this.

recipe/meta.yaml
```diff
@@ -1,69 +1,123 @@
...
-  skip: true  # [win]  # Skip all windows for now due to numpy/scipy version pinning issues
-  skip: True  # [win32]
-  skip: True  # [win64 and py<35]
-  skip: True  # [osx and py34]
-  skip: True  # [py26]
-  skip: True  # [py33]
+  number: 1
+  skip: True  # [not x86_64]
+  skip: True  # [win and vc<14]
+  script: "{{ PYTHON }} -m pip install --no-deps -vv {{ fn }}"  # [not linux]
```
ocefpaf: This does not work and fails with:

```
Found a build.sh script and a build/script section inside meta.yaml.
Either remove the build.sh script or remove the build/script section in meta.yaml.
```

I have a vague memory of ``conda-build`` allowing this in the past, and not sure what are the alternatives. Multiple outputs have the `install_script` and single output have `script` and `build script`, but we don't have an unified way to use those together as far as I know.


@jjhelmus any suggestions?

jjhelmus: One idea is to remove build.sh completely and split it out into `build_linux.sh` and `build_osx.sh`.  Then the script section can have a different entry for each platform.

recipe/meta.yaml
```diff
@@ -1,69 +1,125 @@
...
-  skip: True  # [win64 and py<35]
-  skip: True  # [osx and py34]
-  skip: True  # [py26]
-  skip: True  # [py33]
+  number: 1
+  skip: True  # [not x86_64]
+  skip: True  # [win and vc<14]
+  script:
+    - "{{ PYTHON }} -m pip install --no-deps -vv {{ fn }}"  # [not linux]
+    - build_linux.sh  # [linux]
```
ocefpaf: @jjhelmus I'm guessing this is not the syntax :-(

jjhelmus: Right, that only works with multiple outputs.  Maybe have the `build.sh` copy + execute $RECIPE_DIR/build_linux.sh and $RECIPE_DIR/build_osx.sh?

ocefpaf: > Right, that only works with multiple outputs.

OK. So my impression of this was correct.

> Maybe have the `build.sh` copy + execute $RECIPE_DIR/build_linux.sh and $RECIPE_DIR/build_osx.sh?

The problem with that is that I do not have the recipe's `fn` variable, for the wheel file in OS X, in the build script.

hajapy: What if you write the contents out to a file in the script before calling `build_x.sh`?

`echo "TF_FILENAME={{ fn }}" > properties.txt` and from within `build_x.sh`, do `source properties.txt`.

ocefpaf: That may work but I'd rather have a less hacky solution. Looks like this is a limitation in conda-build and maybe we should try to fix it instead.

ocefpaf:
The OS X install seems to be hanging :-/

recipe/meta.yaml
hajapy: Would it be less hacky enough to just invoke your script with an env far passed in on the script line 
`TF_FILENAME={{ fn }} ./build-custom.sh`?

Waiting for a new feature in conda-build (eg. defining custom env vars available in scripts or making jinja template variable available in scripts) seems like it could be a while...

ocefpaf: @jjhelmus your suggestion worked! Thanks!

jjhelmus: Woohoo!  LGTM, I'll merge once the Linux builds pass on CI.

ocefpaf:
:tada: all green!

jakirkham:
Wow! Thanks @ocefpaf! This is really great to see. 😄

PR #86, opened 2019-09-07
fix libtensorflow_cc headers

njzjz:
resolves #83 

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #87, opened 2019-09-08
correct absl include directory (following #86)

njzjz:
following #86
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/cp_libtensorflow_cc.sh
+++ b/recipe/cp_libtensorflow_cc.sh
@@ -8,6 +8,7 @@ find bazel-genfiles/ -name "*.cc" -type f -delete
 find tensorflow/cc -name "*.cc" -type f -delete
 find tensorflow/core -name "*.cc" -type f -delete
 find third_party -name "*.cc" -type f -delete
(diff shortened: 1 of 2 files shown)


hajapy:
I think it would be very worthwhile to create simple test examples using the c and c++ apis. Right now we're finding issues with the built packages a little bit too late. Something as simple as a hello world, getting the version, and doing a simple arithmetic. Can pull something out of the upstream project too: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api_test.cc

hajapy:
Also, does this mean the prior builds should be moved to the broken label?

njzjz:
> Also, does this mean the prior builds should be moved to the broken label?

Adding a test is a pretty good idea.

Yes I tried to compile a program with libtensorflow_cc but the include files were not correct. TensorFlow doesn’t give a official api for c++ so I have to select headers on my own.

PR #138, opened 2021-10-11
WIP: libprotobuf 3.16

hmaarrfk:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -185,7 +185,8 @@ outputs:
       build:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
-        - bazel
+        - bazel 3        # [not (osx and arm64)]
+        - bazel >=4.2.1  # [osx and arm64]
       host:
         - python
         - pip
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Official builds seem to be using 3.16 for protobuf: https://github.com/tensorflow/tensorflow/blob/v2.6.0/tensorflow/tools/ci_build/install/install_pip_packages.sh#L54

hmaarrfk:
I also have a feeling that grpc-cpp could be unpinned. Maybe.
https://github.com/tensorflow/tensorflow/blob/v2.6.0/tensorflow/tools/ci_build/release/requirements_common.txt

We could also migrate libprotobuf after this passes.

hmaarrfk:
@xhochy it seems you are right. This is REALLY strange though, this this function exists in libprotobuf 3.16 and 3.15. 

I'm just worried that if we don't unpin, we will be unable to install this with newer packages that are only built with libprotobuf 3.18.

hmaarrfk:
Personal debugging information:

1. It seems that tensorflow includes the file `util/type_resolver.h` to eventually include the `Status` class.
```
$ grep "include.*json_util.h" . -R | grep prot
./tensorflow/core/platform/protobuf.h:#include "google/protobuf/util/json_util.h"
```
2. It seems that the only difference between the 3.15 and the 3.16 header is an accidental re-inclusion of the `stubs/status.h` header.
image

However, it seems they used a new namespace.

I think I found the patch that uses the "new" behavior. Trying it now.

PR #142, opened 2021-10-18
Unpin libprotobuf and grpcio

hmaarrfk:
Closes https://github.com/conda-forge/tensorflow-feedstock/pull/139 but with more streamlined git history.

Closes https://github.com/conda-forge/tensorflow-feedstock/issues/135

Linux Builds:
https://github.com/conda-forge/tensorflow-feedstock/pull/142#issuecomment-946469393
https://github.com/conda-forge/tensorflow-feedstock/pull/142#issuecomment-947190800

OSX Builds:
https://github.com/conda-forge/tensorflow-feedstock/pull/142#issuecomment-949307359

Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #147, opened 2021-10-20
[Wip] Cuda home

DerThorsten:
try to build with explicit set coda home


hmaarrfk:
Likely probably have to update:
image


As well. Maybe just the environment variable CUDA_HOME?

Also, I'm much happier with: https://github.com/conda-forge/tensorflow-feedstock/pull/142 as a starting point. It is a cleaner recipe and adds a test.

DerThorsten:
@hmaarrfk I am mostly interested in the CUDA build. Would you still prefer  #142  in that case?

PR #192, opened 2021-12-09
WIP: 2.7.0 Redone

hmaarrfk:
Trying to refactor a few PRs made to the tensorflow 2.7.0 attempt
Checklist

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -63,3 +63,8 @@ bazel %BAZEL_OPTS% build %BUILD_OPTS% %BUILD_TARGET%
 mkdir -p %SRC_DIR%\\tensorflow_pkg
 bazel-bin\\tensorflow\\tools\\pip_package\\build_pip_package %SRC_DIR%\\tensorflow_pkg
 
+:: install the whl using pip
+pip install --no-deps %SRC_DIR%\\tensorflow_pkg\\*.whl
+
+:: The tensorboard package has the proper entrypoint
+rm -f %LIBRARY_BIN%\\tensorboard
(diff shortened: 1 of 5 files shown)


h-vetinari:
So this seems to be missing at least the commits that actually bump to 2.7.0 😅 

Also what's the idea behind the commit here? Putting all the dependencies into the global requirements, rather than per-output? While that should have the same result, I think it's a bit misleading, so I'm not the biggest fan.

xhochy:
With outputs, we shouldn't have a name that matches an output. This triggers a lot of odd behaviours in conda-build, I had quite a pain with that in the past.

hmaarrfk:
i guess i was trying to take advantage of the oddities.

I really don't like the duplication of the dependencies.

but the goal here is to cleanup the dependencies, before moving to 2.7. i appreciate the feedback.

for example, in 2.6.2 i unpinned a dependency on `split` but not on base, so it didn't really do anything.

xhochy:
I don't think we even need the top-level dependencies, we can possibly drop them.

hmaarrfk:
the `build.sh` script does some heavy lifting. do you want to push to this branch to demo the reorg you are suggesting?

xhochy:
> the `build.sh` script does some heavy lifting. do you want to push to this branch to demo the reorg you are suggesting?

Ah, damn. I forgot that we also have `libtensorflow*` outputs 🤦

hmaarrfk:
I think i got the pip command wrong. Would you prefer strange `conda-build` behavior, or fancy jinja syntax?

hmaarrfk:
Fancy jinja syntax used:
https://github.com/conda-forge/tensorflow-feedstock/pull/196

PR #213, opened 2022-03-01
v2.8.0 with system LLVM

### LINKED ISSUE #155: [RFC] unbundle LLVM?
Not sure how realistic this is, but the compilation of this package is spending _a lot_ of time to recompile llvm/lib, which we're already doing in https://github.com/conda-forge/llvmdev-feedstock. There's other parts coming from MLIR, where I haven't check if we're building that separately somewhere.

Not sure if the surgery for this is feasible TBH, but thought it might be worth discussing. There's much more to (potentially) unbundle, see the upstream [third_party](https://github.com/tensorflow/tensorflow/tree/master/third_party) folder... This also shows another potential problem, because tensorflow vendoring LLVM means that they're not necessarily in sync with (the ABI of) released versions.

CC @xhochy @isuruf @hmaarrfk

xhochy:
Checklist

Closes #155


xhochy:
This is a draft to use pre-built binaries for `llvm` and `mlir`. This depends on the following PRs:

- https://github.com/conda-forge/llvmdev-feedstock/pull/148
- https://github.com/conda-forge/mlir-feedstock/pull/12
- https://github.com/conda-forge/openmp-feedstock/pull/59

I created a script called `generate_llvm.py` that generates the Bazel configuration for LLVM from the one that is part of the LLVM codebase. Sadly my efforts were only partially successful as I had a lot of missing symbols at the end. Thus this include the "hack" that the `Support` libraries link in all static libraries and hope that the linker does `--as-needed` linkage correctly. If we want to keep this hack, I can probably trim the Python script a bit more down than what has been needed for the non-hacky version.

@h-vetinari @hmaarrfk @ngam @isuruf You're probably the right audience to look at this and assess whether it would be interesting to push it over the finish line.

xhochy:
Locally on `osx-*`, this seems build successfully 💚

h-vetinari:
> This is a draft to use pre-built binaries for `llvm` and `mlir`.

This is _amazing_. ❤️ 

I don't yet understand the symbol and bazel config issues, but I don't see why we shouldn't merge the precursor PRs and continue iterating here; though perhaps with the one caveat that if we think this has a reasonable chance of working, then I'd prefer to have a dedicated label `llvm_tf` (instead of `llvm_dev`) and builds à la: `14.0.0.tf28` or similar.

hmaarrfk:
Are you able to add the dev packages to a channel and ask Azure to build them? Does this get the builds completed in time remotely?

isuruf:
You can have a look at https://github.com/conda-forge/tensorflow-feedstock/pull/78/files as well.

`--as-needed` doesn't work with static libraries.

xhochy:
> You can have a look at https://github.com/conda-forge/tensorflow-feedstock/pull/78/files as well.
> 
> `--as-needed` doesn't work with static libraries.

Yes, the PR was my initial inspiration but sadly the LLVM Bazel definitions got a lot more complicated and diverged from what is built with CMake.

As `--as-needed` won't work with static libraries, I'll build my own version of it then 🤣

xhochy:
> > This is a draft to use pre-built binaries for `llvm` and `mlir`.
> 
> This is _amazing_. ❤️
> 
> I don't yet understand the symbol and bazel config issues, but I don't see why we shouldn't merge the precursor PRs and continue iterating here; though perhaps with the one caveat that if we think this has a reasonable chance of working, then I'd prefer to have a dedicated label `llvm_tf` (instead of `llvm_dev`) and builds à la: `14.0.0.tf28` or similar.

Would be nice to discuss that over at https://github.com/conda-forge/llvmdev-feedstock/pull/148 I would also use these binaries for `jaxlib` but in the end that is also Tensorflow which actually builds it. Once this is discussed and merged, we can get the CI started here.

recipe/test_libtensorflow_cc.sh
```diff
@@ -2,5 +2,8 @@
 
 set -exuo pipefail
 
-$CXX -std=c++14 -o test_cc -L${PREFIX}/lib/ -ltensorflow_cc -ltensorflow_framework -lrt -I${PREFIX}/include/ test_cc.cc
+if [[ "${target_platform}" == linux-* ]]; then
+  export LD_FLAGS="${LDFLAGS} -lrt"
```
isuruf, line 6:
```suggestion
  export LDFLAGS="${LDFLAGS} -lrt"
```

xhochy:
With all the changes, it builds in `Total time: 1:38:57.4` (osx-*-py310) locally on my first-generation M1. It will be interesting to see what this means in CI time.

ngam:
> With all the changes, it builds in `Total time: 1:38:57.4` (osx-*-py310) locally on my first-generation M1. It will be interesting to see what this means in CI time.

@xhochy can I help in any way to test things here? I am sorry I've been absent

Btw, I've been thinking about modifying the docker calls to utilize singularity; this way, I can potentially build this on super powerful machines, but it is really tricky because singularity doesn't ensure isolation by default like docker does...

xhochy:
Doesn't yet lead to positive results 🤷 

on master it ends with:

```
[11,645 / 17,178] Compiling tensorflow/compiler/mlir/tensorflow/transforms/tpu_sharding_identification_pass.cc; 21s local ...
```

here it ends with:

```
[5,838 / 6,248] Compiling tensorflow/core/kernels/linalg/einsum_op_impl_bfloat16.cc; 48s local ...
```

Let's see whether Linux provides better results.

xhochy:
> @xhochy can I help in any way to test things here? I am sorry I've been absent

We will need to see what CI runtimes tell us. The build on `osx-64` thinks internally in bazel that it is a cross-compilation where it clearly shouldn't be but I don't know how to tell it explicitly not to do that. I also expect nobody else to figure it out.

recipe/conda_build_config.yaml
```diff
@@ -3,3 +3,23 @@
 # available on 10.13
 MACOSX_SDK_VERSION:        # [osx and x86_64]
   - "10.13"                # [osx and x86_64]
+c_compiler_version:        # [linux]
```
hmaarrfk, line 6: Now that things seem working. Can we try to remove this? I applied a GCC patch to Fonda forge a few days back

xhochy: I will do this once the linux builds pass again with GCC 9. I have introduced some kind of issue that I want to figure out before making a change there again.

recipe/test_libtensorflow.sh
```diff
@@ -2,5 +2,5 @@
 
 set -euxo pipefail
 
-$CC -o test_c -L${PREFIX}/lib/ -ltensorflow -I${PREFIX}/include/ test_c.c
+${CC} -o test_c ${CFLAGS} ${LDFLAGS} -ltensorflow test_c.c
```
hmaarrfk, line 5: This is expanding to:
```
/bin/x86_64-conda-linux-gnu-cc \
-o test_c \
-march=nocona -mtune=haswell -ftree-vectorize -fPIC \
-fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe \
-isystem /home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566 \
/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/include -fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566/test_tmp=/usr/local/src/conda/libtensorflow-2.8.0 \
-fdebug-prefix-map=/home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold=/usr/local/src/conda-prefix \
-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections \
-Wl,--allow-shlib-undefined \
-Wl,-rpath,/home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib -Wl,-rpath-link,/home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib \
-L/home/conda/feedstock_root/build_artifacts/tensorflow-split_1647631094566/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib \
-ltensorflow test_c.c
```

Seems to be correct.

xhochy: Somehow the order mattered locally.

hmaarrfk:
Should we start building more thoroughly? Or do you want to huntdown the bug that is causing the strangeness with `CFLAGS`

hmaarrfk:
Seems like a fix is proposed:
https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/68

xhochy:
@conda-forge-admin[](https://github.com/xhochy) please rerender

xhochy:
@conda-forge-admin please rerender

h-vetinari:
Awesome that you unbundled oneDNN as well. It's already really cool to have the CPU builds run through with ample time to spare!

xhochy:
FTR: the latest commit took `582,69s user 589,93s system 13% cpu 2:29:32,04 total` for `conda mambabuild -m .ci_support/osx_arm64_python3.10.____cpython.yaml -c conda-forge/label/llvm_tf -c conda-forge/label/mlir_tf recipe` on my M1 using an x86 mambaforge.

xhochy:
@conda-forge-admin please rerender

xhochy:
@conda-forge-admin please rerender

xhochy:
@conda-forge/tensorflow I would like to build this (locally). Could I get a review before I spent the CPU hours?

recipe/generate_llvm.py
```diff
@@ -0,0 +1,371 @@
...
+exports_files(
+    {files}
+)
+"""
+
+
+# Generate LLVM build defintion
+
+PREFIX = Path(os.environ["PREFIX"])
+SYMBOL = "LLVM"
```
h-vetinari, line 128: Maybe I'm overlooking something, but I find this a strange variable name. In the loop above you're really resolving symbols from the `nm` output, but calling a variable containing the string `"LLLVM"` a symbol is confusing me.

recipe/generate_llvm.py
```diff
@@ -0,0 +1,371 @@
...
+    "RISCV",
+    "SystemZ",
+    "X86",
+]
+
+def select(*args, **kwargs):
+    return []
+
+# Load the BUILD file and run it with the above defined custom functions
+exec(r.text)
```
h-vetinari, line 231: Not a big fan of the distance between creating `r` and using it, especially given it's very short name.

Can we do `upstream_build_bazel = r.text` after the `r=` above and then 
```suggestion
# Run the upstream BUILD file with the above defined custom functions
exec(upstream_build_bazel)
```
here?

recipe/generate_llvm.py
```diff
@@ -0,0 +1,371 @@
...
+def if_cuda_available(*args, **kwargs):
+    return []
+
+package = load = filegroup = empty
+genlinalg = cc_library
+# TODO: CAPI
+mlir_c_api_cc_library = cc_library
+
+# Load the BUILD file and run it with the above defined custom functions
+exec(r.text)
```
h-vetinari, line 364: Same here (even larger distance between definition and use)

recipe/meta.yaml
```diff
@@ -65,16 +74,19 @@ requirements:
...
     # unpinned; though some restrictions come in for packages
     # that are also listed as a requirement for the pip_package.
+    - abseil-cpp
     - astor
     - cython
     - dill
     - giflib
     - grpc-cpp
-    - flatbuffers
+    - flatbuffers <2.0.6
```
h-vetinari, line 83: Can we have a comment for the upper limit?

recipe/meta.yaml
```diff
@@ -253,7 +270,7 @@ outputs:
         # decide on the cuda variant
         # https://github.com/conda-forge/tensorflow-feedstock/issues/162
         - {{ compiler('cuda') }}                 # [cuda_compiler_version != "None"]
-        - bazel
+        - bazel <5
```
h-vetinari, line 273: Above we have `bazel 4.2.1`, here `bazel <5`; Is there a reason for the difference?

recipe/patches/0002-Switch-to-system-llvm-project.patch
```diff
@@ -0,0 +1,61 @@
...
++	build_file_content = "# empty",
++    )
++    #tf_http_archive(
++    #    name = "llvm-project",
++    #    build_file = "//third_party/llvm:BUILD",
++    #    sha256 = "22054926b426c66d8f2bc22071365df6e35f3aacf19ad943bc6167d4cae3bebb",
++    #    strip_prefix = "no_valid_prefix",
++    #    system_build_file = "//third_party/systemlibs:llvm.BUILD",
++    #    urls = tf_mirror_urls("https://github.com/invalid/mirror/source.tar.gz"),
++    #)
```
h-vetinari, line 40: This commented out code doesn't seem to have existed previously - any reason we're patching it in?

h-vetinari (commented):
Did a slightly-less-than-superficial pass, looking very good. Most of my comments are on the level of nits and not critical.

recipe/conda_build_config.yaml
```diff
@@ -3,3 +3,5 @@
 # available on 10.13
 MACOSX_SDK_VERSION:        # [osx and x86_64]
   - "10.13"                # [osx and x86_64]
+channel_sources:
+  - conda-forge/label/mlir_tf,conda-forge/label/llvm_tf,conda-forge
```
hmaarrfk, line 7: I already forgot what `mlir` stood for. Could you add a sentence or two as to what these are for.

recipe/gen-bazel-toolchain.sh
```diff
@@ -38,14 +45,19 @@ pushd custom_toolchain
         -e "s:\${target_platform}:${target_platform}:" \
         -e "s:\${INSTALL_NAME_TOOL}:${INSTALL_NAME_TOOL}:" \
         -e "s:\${CONDA_BUILD_SYSROOT}:${CONDA_BUILD_SYSROOT}:" \
+	-e "s:\${MACOSX_SDK_VERSION}:${MACOSX_SDK_VERSION:-}:" \
+	-e "s:\${MACOSX_DEPLOYMENT_TARGET}:${MACOSX_DEPLOYMENT_TARGET:-}:" \
```
hmaarrfk, line 49:
```suggestion
	    -e "s:\${MACOSX_SDK_VERSION}:${MACOSX_SDK_VERSION:-}:" \
	    -e "s:\${MACOSX_DEPLOYMENT_TARGET}:${MACOSX_DEPLOYMENT_TARGET:-}:" \
```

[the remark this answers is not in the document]
hmaarrfk: I think it is correct. for some reason the tab character isn't rendering correclty.

recipe/meta.yaml
hmaarrfk, line 83:
```suggestion
    # the latest flatbuffers 2.0.6 seems to have removed a sybmol in a patch release...
    # https://github.com/google/flatbuffers/issues/7217
    - flatbuffers <2.0.6
```

hmaarrfk (commented):
This is seriously complicated. 

Can we have a summary as to why this is useful? We don't get our builds under the CI limit for the GPU builds.

It doesn't seem like these changes can be upstreamed.

I fear like this is simply not going to be maintainable.

xhochy:
> Can we have a summary as to why this is useful? We don't get our builds under the CI limit for the GPU builds.

Getting GPU or OSX builds running in CI would have been something that would have offset the maintenance costs here. Without that I feel I have nothing in my hands to argue why this should be maintained.

recipe/meta.yaml
```diff
@@ -253,7 +270,7 @@ outputs:
...
         - bazel >=4.2.1  # [osx and arm64]
```
xhochy, line 274:
```suggestion
```

xhochy:
> Without that I feel I have nothing in my hands to argue why this should be maintained.

Still, I won't give up here.

PR #220, opened 2022-04-27
Icu70 and 2.7.1

hmaarrfk:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #221, opened 2022-04-27
Rerender 2.7.0

xhochy:
Checklist


xhochy:
@hmaarrfk This builds locally. While only being the most minimal improvement, it at least gets us forward by a bit. Given that we have been struggling with the "big bang", I would go step-by-step and volunteer to build quite a bit.

hmaarrfk:
ok. I also made:
https://github.com/conda-forge/tensorflow-feedstock/pull/220
for ICU70 and 2.7.1

It just takes a full day of a machine to build tensorflow for GPU. Not sure if we are ready to drop many cuda builds. That would make things much faster.

hmaarrfk:
For your build logs. it seems that the compiler downgrade is also necessary.

ngam:
> It just takes a full day of a machine to build tensorflow for GPU. Not sure if we are ready to drop many cuda builds. That would make things much faster.

Just a minor suggestion: for this interim building, could we target one or two specs of each? I think building py39 and py310 (or just py310) for osx-64, osx-arm64, linux-64 (no cuda), and linux-64 (cuda112). That will make it only 4 or 8 builds only. Once we feel we have something good working, we can then build all for only major releases (i.e.  I would build all only for 2.8.0 and 2.9.0 tbh)

In other words, drop py37 and py38 as well as cuda<112 for sure. Consider dropping either py39 or py310 as well.

h-vetinari:
> In other words, drop py37 and py38 as well as cuda<112 for sure. Consider dropping either py39 or py310 as well.

I'd rather not drop python versions, because this might lead to weird solver errors. I'd be fine with dropping CUDA<11.2

hmaarrfk:
Ultimately. Once we have. A build going i can build Linux fine. But I just can't get cuda builds to work.

hmaarrfk:
Thank you xhochy for trying a different approach to the way I'm going about it. Happy to build whatever will help.

xhochy:
Seems like this is passing. I'll run the local builds later today and report back.

hmaarrfk:
Thank you for all the good work. I've pulled most of your stuff with 2.7.1 and icu70 in https://github.com/conda-forge/tensorflow-feedstock/pull/220 

I'm starting a build of cuda 11.2 and python 3.9 on linux. I have high hopes.

Issue #223, opened 2022-04-29
Not a really a bug, but let's enable XLA in the builds going forward


ngam:
### Solution to issue cannot be found in the documentation.

- [X] I checked the documentation.

### Issue

```
2022-04-29 12:25:31.051195: W tensorflow/compiler/jit/kernels/xla_ops.cc:436] Compilation failed:UNIMPLEMENTED: Could not find compiler for platform CUDA: NOT_FOUND: could not find registered compiler for platform CUDA -- check target linkage (hint: try adding tensorflow/compiler/jit:xla_gpu_jit as a dependency).  Falling back to TF function call.
```

This should be 1, any reason why it is not?
https://github.com/conda-forge/tensorflow-feedstock/blob/c43f2820986d3eb74f63c23eadf85313dc54c635/recipe/build.sh#L97

### Installed packages

```shell
n/a
```

### Environment info

```shell
n/a
```

ngam:
More from NVIDIA's build for us to follow if possible:

```bash

export TF_NEED_CUDA=1
export TF_NEED_CUTENSOR=1
export TF_NEED_TENSORRT=1
export TF_CUDA_PATHS=/usr,/usr/local/cuda
export TF_CUDA_VERSION=$(echo "${CUDA_VERSION}" | cut -d . -f 1-2)
export TF_CUBLAS_VERSION=$(echo "${CUBLAS_VERSION}" | cut -d . -f 1)
export TF_CUDNN_VERSION=$(echo "${CUDNN_VERSION}" | cut -d . -f 1)
export TF_NCCL_VERSION=$(echo "${NCCL_VERSION}" | cut -d . -f 1)
export TF_TENSORRT_VERSION=$(echo "${TRT_VERSION}" | cut -d . -f 1)
export TF_ENABLE_XLA=1
export TF_NEED_HDFS=0
if [ "${TARGETARCH}" = "amd64" ] ; then export CC_OPT_FLAGS="-march=sandybridge -mtune=broadwell" ; fi
if [ "${TARGETARCH}" = "arm64" ] ; then export CC_OPT_FLAGS="-march=armv8-a" ; fi
export TF_USE_CCACHE

# This is the flags that Google use internally with clang.
# Adding them help make the compilation error more readable even if g++ doesn't do exactly as clang.
# Google use -Werror too. But we can't use it as the error from g++ isn't exactly the same and will trigger useless compilation error.
[...]

PR #226, opened 2022-05-01
WIP: Just try to rebuild the old stuff the exact same way as we did before.

### LINKED ISSUE #225: @conda-forge-admin please rerender
### Additional comment:

_No response_

conda-forge-linter:
hmaarrfk: I'm trying to rebuild this exactly as we had it in the first build. For some reason, it is giving me real trouble.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,3 +344,6 @@ Feedstock Maintainers
 * [@wolfv](https://github.com/wolfv/)
 * [@xhochy](https://github.com/xhochy/)
 
+
+<!-- dummy commit to enable rerendering -->
+
--- a/recipe/custom_toolchain/cc_toolchain_config.bzl
+++ b/recipe/custom_toolchain/cc_toolchain_config.bzl
@@ -311,6 +311,7 @@ def _impl(ctx):
 
         if (len("${CUDA_HOME}")):
             cxx_builtin_include_directories.append("${CUDA_HOME}/include")
+            cxx_builtin_include_directories.append("${CUDA_HOME}/targets/x86_64-linux/include/")
 
     return cc_common.create_cc_toolchain_config_info(
         ctx = ctx,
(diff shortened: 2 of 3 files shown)


hmaarrfk:
@jakirkham did anything change in the cuda images? I can't seem to build this feedstock again, with what I believe are the the same settings as last time.

ngam:
> @jakirkham did anything change in the cuda images? I can't seem to build this feedstock again, with what I believe are the the same settings as last time.

Also copying @leofang and @kkraus14 who were helpful in discussing the cuda override issue.

@hmaarrfk if it's not a hassle, could you please put the error you get in the comments? Thanks for working on this!!

ngam:
(as in, are they the same as in the logs here or are they different?)

hmaarrfk:
A few packages are the same, but I didn't rerender :/ I can try to pin packages a little harder, but that is more time than I have for today.

ngam:
:/ sorry this is turning into a nightmare :/

kkraus14:
> @jakirkham did anything change in the cuda images? I can't seem to build this feedstock again, with what I believe are the the same settings as last time.

The only thing that I know of that have changed recently is a PR I pushed of changing the CUDA headers to be system includes instead of user includes: https://github.com/conda-forge/nvcc-feedstock/pull/80

You could try pinning to the previous build of `nvcc_linux-64` to see if that resolves your issue. That being said, if that fixes your issue that's pointing towards something problematic in the build here.

hmaarrfk:
Ok. That might actually be the issue. It says we aren't depending on the right things

hmaarrfk:
THank you @kkraus14 reverting that that seemed to work. Now I think xhochy  is smarter than me and may have found a better way. So i'm going to try his fix too.

ngam:
@hmaarrfk are we all good now?

Btw, do you have any thoughts on enabling xla and including (if not already) the nvcc compiler as a runtime dep? (ref #223)

PR #237, opened 2022-05-16
RFC: Build libtkernel_all_kernels as standalone package

xhochy:
Checklist


xhochy:
If this doesn't make it in time, I would have a look at separating out `libtensorflow_framework`

hmaarrfk:
Is the idea to have a separate package for `libtkernel_all_kernels`? not built in this feedstock?

ngam:
Title says "RFC", my C: 100% support if we can actually combine this smoothly later with the rest. 

Even if this effort fails, this PR is really useful for a slightly modified approach. I've been exploring the strategy of staggered build inside containers:

1. build to a certain stage (e.g. this PR)
2. save container and publish it as is (e.g. through ghcr.io)
3. continue where we left off in a subsequent job (i.e. pull container, `FROM x` and keep going)

etc.

This will require some fundamental edits to the workflow as governed by conda-smithy, but we can also pull it of I believe. We  can have an exemption for select packages like tensorflow where there a staggered build approach is warranted. But if this works out-of-the-box like you seem to be suggesting, it will be massive game changer!!

ngam:
Title says "RFC", my C: 100% support if we can actually combine this smoothly later with the rest. 

Even if this effort fails, this PR is really useful for a slightly modified approach. I've been exploring the strategy of staggered build inside containers:

1. build to a certain stage (e.g. this PR)
2. save container and publish it as is (e.g. through ghcr.io)
3. continue where we left off in a subsequent job (i.e. pull container, `FROM x` and keep going)

etc.

This will require some fundamental edits to the workflow as governed by conda-smithy, but we can also pull it of I believe. We  can have an exemption for select packages like tensorflow where there a staggered build approach is warranted. But if this works out-of-the-box like you seem to be suggesting, it will be massive game changer!!

xhochy:
> Is the idea to have a separate package for `libtkernel_all_kernels`? not built in this feedstock?

Yes, `libtensorflow_framework` and `libtkernel_all_kernels` are both shared libraries that describe a quite large subset of the tensorflow packages. Sadly, it seems the kernels are a really large subset and thus they still don't succeed in 6h for CUDA. Thus this only brings us a small benefit if it works (only 4 manual builds).

hmaarrfk:
What about combining this with your system llvm build? seems like it might be enough

xhochy:
It would have been nice without it. 🤷

hmaarrfk:
Needs one more: https://github.com/conda-forge/cyvlfeat-feedstock/pull/10

xhochy:
> Needs one more: [conda-forge/cyvlfeat-feedstock#10](https://github.com/conda-forge/cyvlfeat-feedstock/pull/10)

I don't understand that cross-reference.

hmaarrfk:
sorry, wrong github chat for me.

hmaarrfk:
> It would have been nice without it. shrug

I get it, but ultimately, "one more step" in a multi-step build process isn't going to kill us.

It is rather hard to coordinate all the people building things, and we still haven't started to build windows...

xhochy:
I've tested this with the LLVM unbundling and it also doesn't get below 6h. Thus, this won't bring us any further. There are two possible approaches:

- We find a smaller intermediate that gets us below the 6h mark also for CUDA buils.
- We use libtfkernels_all_kernels as an intermediate result and only need to do manual builds for it and can built the rest of tensorflow in CI. This would at least improve the current situation and would make manual builds much leaner.

hmaarrfk:
Whats the probability of splitting `all_kernels_impl` into a few static libraries that we can compile as distinct outputs (cpu, gpu) and then combining them together?

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L623

hmaarrfk:
It might need a patch like:
```
--- tensorflow/core/BUILD.original      2022-05-17 22:30:51.668034914 -0400
+++ tensorflow/core/BUILD       2022-05-17 22:37:49.352329569 -0400
@@ -700,7 +700,12 @@
         "//tensorflow/core/kernels/mkl:mkl_matmul_op",
         "//tensorflow/core/kernels/mkl:mkl_tfconv_op",
         "//tensorflow/core/kernels/mkl:mkl_tmp_bf16_ops",
-    ]) + if_cuda_or_rocm([
+    ])
+)
+cc_library(
+    name = "all_kernels_impl_hardware",
+    visibility = [":__subpackages__"],
+    deps = if_cuda_or_rocm([
         "//tensorflow/core/kernels:cudnn_rnn_kernels",
     ]) + if_cuda([
         "//tensorflow/core/grappler/optimizers:gpu_swapping_kernels",
@@ -710,7 +715,7 @@
     ]) + if_tensorrt([
         "//tensorflow/compiler/tf2tensorrt:trt_engine_resource_op_kernels",
         "//tensorflow/compiler/tf2tensorrt:trt_op_kernels",
-    ]) + tf_tpu_dependencies(),
+    ]) + tf_tpu_dependencies()
 )

 cc_library(
```
then we would need to add the output `all_kernels_impl_hardware` as a dependency to the library (maybe).

h-vetinari:
> Whats the probability of splitting `all_kernels_impl` into a few static libraries that we can compile as distinct outputs (cpu, gpu) and then combining them together?
> 
> https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/BUILD#L623

There's some more layers to deal with from `all_kernels_impl` to `all_kernels`, but fundamentally I agree, and I think that the list of dependencies in `tf_tpu_dependencies` would lend itself well to splitting (e.g. core / mkl / hardware**) without introducing undue maintenance overhead.

** might be even split further depending on which builds are the biggest.

xhochy:
From local testing, there seem to be some-but-few very compile-intensive kernels. There is already machinery to build for each kernel a `libtfkernel_<kernel-name>.so`. I would build that out for some of the biggest libraries and then replace them with the respective shared libraries.

As all of them depend on `libtensorflow_framework`, I would first try to separate that out into a standalone package before moving on to splitting of kernels. This will also verify whether you actually can separate out a lot of the work.

xhochy:
Closing in favor of #247

PR #240, opened 2022-05-17
v2.9.0

ngam:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #248, opened 2022-05-29
RFC: libtfkernel_unique_op

xhochy:
Checklist


xhochy:
This also builds `conv_2d_gpu_uint8` which I wouldn't have expected from the dependency graph.

h-vetinari:
This looks pretty good! I see you had a comment about MLIR dependence of `unique_op` but that seems to have been deleted? I guess that means we're using the TF-bundled MLIR?

Do you think it would make sense to bundle the various kernel outputs into a single feedstock an template on the libname with cbc.yaml?

xhochy:
Not sure which comment you are referring to but I'm still in troubles with MLIR over at #249. Before that isn't solved, the splitting doesn't yet provide noticeable benefit. I'm currently in progress of building a different version of pre-build LLVM that might help.

h-vetinari:
> Not sure which comment you are referring to [...]

This one from https://github.com/conda-forge/tensorflow-feedstock/pull/248/commits/5a66943b2d7995ad68d6f03d8490f1969574b0cd:
```
# This is heavily dependent on MLIR-generated kernel code.
# It should probably be its own build with the other MLIR-based ops.
# We could split out the MLIR-based ones into individual ones once we can use a prebuilt MLIR.
#  //tensorflow/core/kernels:libtfkernel_cwise_op.so
#  //tensorflow/core/kernels:libtfkernel_unique_op.so
#  //tensorflow/core/kernels:libtfkernel_dynamic_partition_op.so
```

xhochy:
This is now at https://github.com/conda-forge/tensorflow-feedstock/pull/249/files#diff-d7075654874cb08007a21aaab3ecd4b3453a9087e7505d034d548b8938b599bcR29

ngam:
Should we try to build 2.9.1 in the meanwhile or should we (i.e. xhochy) press with this to see if we can get it to work?

xhochy:
Try to get 2.9.1 to work, I can rebase my patches later. I have still not seen this working end-to-end, so I won't be 100% sure it works.

ngam:
> Try to get 2.9.1 to work, I can rebase my patches later. I have still not seen this working end-to-end, so I won't be 100% sure it works.

I believe the linux-64 builds are ready. I was hoping you could look at the osx errors (or maybe they're only CI related?)

#240

PR #249, opened 2022-05-29
RFC: libtfkernel_mlir_ops

xhochy:
Checklist


xhochy:
This worked 🙌

xhochy:
Sadly loading this doesn't work in a split build as some math operations are loaded twice (which is not allowed). Possibly the best approach is to do the same as we did with LLVM: Build some intermediate static libraries and reuse them in a later build.

ngam:
> Sadly loading this doesn't work in a split build as some math operations are loaded twice (which is not allowed). Possibly the best approach is to do the same as we did with LLVM: Build some intermediate static libraries and reuse them in a later build.

Sorry to hear this.

I am curious if you think we can trim some of the compilations. I do notice that we end up compiling for ROCm and TPUs even though we don't really support those. Or maybe I am misreading the compilation logs?

h-vetinari:
> Possibly the best approach is to do the same as we did with LLVM: Build some intermediate static libraries and reuse them in a later build.

This sounds fine to me TBH. Whether the intermediate libs are static or dynamic doesn't really matter IMO.

PR #252, opened 2022-06-10
Build with tensorflow-build-cache

xhochy:
Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,86 +8,106 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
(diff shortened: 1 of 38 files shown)


ngam:
Two quick comments:
- we may need either remote caching (i.e. to a google cloud bucket or something) or some way to reuse CI caching. If you have thought about this enough, I am happy to start a bucket for our use here to see if it works.
- Either way though, we need to figure out a way to bypass the first build --- when I build locally with singularity (which caches the whole thing since it has to be done on the host and not inside the container, i.e. the container space doesn't really get used except as a stationary artifact for all the tools it contains), the build time is reduced by ~50% after the first build (in fact, it also caches the whole conda env, thus speeding through the whole annoying cudatoolkit installation, etc.)

In summary, 
1. to speed up local builds, we can just use singularity (I had a PR, but it needs further work)
2. to speed up CI builds, our likely only solution is remote caching to a google cloud bucket.

xhochy:
> to speed up CI builds, our likely only solution is remote caching to a google cloud bucket.

The approach here uses a conda package as the cache, so no need for any cloud bucket.

PR #260, opened 2022-08-05
2.10.0.rc1

ngam:
notable changes: 
- use latest bazel
- update flatbuffers
- update custom_toolchain (see more potential updates in #261)
- [ ] fix protobuf issues, pin to lower than 3.20, non-inclusive, i.e. <3.20

---

MAJOR CHANGES:

: RIP cuda102 (I am not going to bother looking into this)

```
Error in fail: The following libraries cannot be linked either statically or dynamically:
@cub_archive//:cub
To ignore which libraries get linked statically for now, add the following to 'static_deps':
        "@cub_archive//:__subpackages__",

```

---

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 source:
   - url: https://github.com/tensorflow/tensorflow/archive/refs/tags/v{{ version.replace(".rc", "-rc") }}.tar.gz
-    sha256: 3bac161442497c1ad5fd44ee7084064c0f5aab2cceb53a34c9345eaaa326a618
+    sha256: f3c61ff3f6baf1a38cc232bbb20288c7f0aea80955af18d746d01774579a5ce8
     patches:
       - patches/0001-loosen-requirements.patch
 


h-vetinari:
The recipe currently contains both:
```
  # We specify a version of abseil_cpp that is compatible with C++17
  skip: true  # [abseil_cpp != '20211102.0']
```
as well as
```
  - name: libtensorflow
    [...]
      host:
        - abseil-cpp 20211102.0
```

which seems contradictory.

> @hmaarrfk @h-vetinari I lost track of what is ready for c++17 and what is not. The cuda11x builds are failing with c++17 errors. As you see below the compiler call includes c++17, but the error is obviously c++17 specific. It could be something else bringing c++14 (or something else) with it?
> 
> Any advice? How to make sure cuda11x builds actually work/compile with c++17?

the abseil shared builds should be C++17 already. Based on the error you pasted, it's that the code itself depends on a C++17 facility ([`std::optional`](https://en.cppreference.com/w/cpp/utility/optional)), but perhaps something in our infra or bazel or ...? is still injecting a C++14 somewhere.

h-vetinari:
Thanks a lot for your work on this! Very timely, as previously. 🙃 

> The conclusion from this: We will be ready once 2.10 is out, but we we need to fine-tune both the custom_toolchain and our deps (e.g. protobuf and libprotobuf). I will leave these PRs (this and #261) open for a bit longer and then I will close them.

Why close them if we can take them as the basis for iterating? What protobuf changes are necessary? C++17 builds?

h-vetinari:
Ugh, symbol re-exporting is just blech. If it's not your lib, don't provide the symbols (vendoring something and using it internally is one thing, but seizing other project's symbol namespaces quite another).

h-vetinari:
The linker not finding the destructor reminds me of problems I had during https://github.com/conda-forge/sentencepiece-feedstock/pull/26, and the stuff that shows up as missing virtual functions on windows.

This [answer](https://stackoverflow.com/a/57504289) might be relevant. Beyond that, I'd use the newest abseil builds, and also try the static ones matching the C++ version used to compile tf

PR #270, opened 2022-09-25
reenable all cudas to examine erros in 11.1

ngam:
Checklist

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -8,38 +8,86 @@ jobs:
     vmImage: ubuntu-latest
   strategy:
     matrix:
-      linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.7.____cpython:
-        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.2cxx_compiler_version10numpy1.20python3.7.____cpython
+      linux_64_c_compiler_version10cuda_compiler_version11.1cudnn8cxx_compiler_version10numpy1.20python3.7.____cpython:
+        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.1cudnn8cxx_compiler_version10numpy1.20python3.7.____cpython
+        UPLOAD_PACKAGES: 'True'
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cuda:11.1
+      linux_64_c_compiler_version10cuda_compiler_version11.1cudnn8cxx_compiler_version10numpy1.20python3.8.____cpython:
+        CONFIG: linux_64_c_compiler_version10cuda_compiler_version11.1cudnn8cxx_compiler_version10numpy1.20python3.8.____cpython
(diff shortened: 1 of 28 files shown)


hmaarrfk:
One thing that seems to have changed is the version of the compiler for 11.2 i'm going to try to downgrade locally

h-vetinari:
> One thing that seems to have changed is the version of the compiler for 11.2 i'm going to try to downgrade locally

How is that relevant? It's part of the normal cf-infra and it works for 11.2, so why would we want to do something differently?

(I could kinda understand that from the POV of investigating the failures for 11.0/11.1, but 11.2 is passing)

hmaarrfk:
Because I patched 10.3 for compatibility and removed the patch when it was presumably included in 10.4
https://github.com/conda-forge/ctng-compilers-feedstock/pull/104/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aL23

for what its worth, local compilation with 10.3.0 shows the same error.

hmaarrfk:
There are also a few notes on the internet about compiler incompatibility:
* https://docs.parasoft.com/display/CPPTEST20211/Updates+in+2020.2
* https://forums.developer.nvidia.com/t/nvc-20-9-fails-to-compile-code-instantiating-any-std-tuple-with-gcc-10-2-on-c-17/160987
* https://bitbucket.org/berkeleylab/upcxx/issues/302/build-failure-for-intel-c-with-std-c-17

If you search for `_UElements` in these, you will find errors that look oddly similar. Maybe we should recompile this particular version of protobuf? Maybe this is a reason to move foward with the migration.

h-vetinari:
> Maybe we should recompile this particular version of protobuf? Maybe this is a reason to move foward with the migration.

I don't understand what you mean here. AFAICT the error is from a combination of nvcc and libstdcxx, How does protobuf come into play? Also, 11.2 is passing, so which failures are we talking about?

hmaarrfk:
Specifically about the failures with 11.1

hmaarrfk:
```
/home/conda/feedstock_root/build_artifacts/tensorflow-split_1664113330988/_build_env/x86_64-conda-linux-gnu/include/c++/10.4.0/tuple(582): error: pack "_UElements" does not have the same number of elements as "_Elements"
          detected during:
            instantiation of "__nv_bool std::tuple<_Elements...>::__nothrow_constructible<_UElements...>() [with _Elements=<google::protobuf::Arena *&, int &, int &, google::protobuf::internal::RepeatedPtrFieldBase::Rep *&>, _UElements=<>]" 
(1633): here
            instantiation of "std::tuple<_Elements &...> std::tie(_Elements &...) noexcept [with _Elements=<google::protobuf::Arena *, int, int, google::protobuf::internal::RepeatedPtrFieldBase::Rep *>]" 
/home/conda/feedstock_root/build_artifacts/tensorflow-split_1664113330988/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/include/google/protobuf/repeated_ptr_field.h(315): here


```

PR #286, opened 2022-11-21
tensorflow 2.11

ngam:
- [x] Linux + CUDA -- http://anaconda.org/cftf-linux/ -- [log.txt](https://github.com/conda-forge/tensorflow-feedstock/files/10242390/log.zip)
- [x] Linux + CPU -- http://anaconda.org/cftf-linux/-- [log.txt](https://github.com/conda-forge/tensorflow-feedstock/files/10242390/log.zip)
- [x] OSX x86 -- https://anaconda.org/pkgw-forge/ -- [logs.txt](https://github.com/conda-forge/tensorflow-feedstock/files/10369219/logs.zip)
- [x] OSX arm -- https://anaconda.org/pkgw-forge/ -- [logs.txt](https://github.com/conda-forge/tensorflow-feedstock/files/10369219/logs.zip)

Checklist


hmaarrfk:
can you build anything from osx? maybe even x86?

h-vetinari:
> If nobody attempts to build, I will have another closer look next week.

Should we attempt to migrate libgrpc 1.50 in the meantime (I could do it)?

1.51 is already out (we could even migrate to that[^1]), and tensorflow is one of the trickiest ones to get on even footing w.r.t. grpc, making it hard to install with other packages.

[^1]: seems we don't need to keep a version "in reserve" for the privatization stuff, as the release quite frequently...

h-vetinari:
Grpc PR arrived: #289 

Should be easy to include. :)

Pity to hear you'll lose access to those compute resources, it was very helpful to have you building/debugging things so regularly ♥️

h-vetinari:

hmaarrfk:
I tried to add a pinning to this. Hopefully it helps.

recipe/meta.yaml
```diff
@@ -204,14 +204,18 @@ outputs:
         - opt_einsum >=2.3.2
         - protobuf >=3.9.2
         - python-flatbuffers >=2
-        - flatbuffers >=2
+        # This can be removed once the dust settles after
+        # the run_export is added to the main and 2.x.x branch
+        # https://github.com/conda-forge/flatbuffers-feedstock/pull/35
+        # https://github.com/conda-forge/flatbuffers-feedstock/pull/36
+        - {{ pin_compatible('flatbuffers', max_pin='x.x') }}
```
hmaarrfk, line 211:
```suggestion
        # https://github.com/conda-forge/flatbuffers-feedstock/pull/37
        - {{ pin_compatible('flatbuffers', max_pin='x') }}
```
loosen pin on flatbuffers

hmaarrfk:
of course. i didn't know how far along you were in the build.

@conda-forge-admin please rerender

hmaarrfk:
Um lol, this was just merged https://github.com/conda-forge/conda-smithy/pull/1686 maybe yo uwant to force push and rerender

hmaarrfk:
That refers to OSX arm64 correct?

xhochy:
@ngam I did definitely compile native and cross for osx-arm64 last time I did compile for tensorflow. I'm a bit overloaded currently and don't know whether I find time to look into Tensorflow this year sadly :( 

The latest error looks though as if you would need `cython` in `build` and link Bazel correctly to it.

hmaarrfk:
IMO. we should just keep tensorflow 2.10 if we can't get OSX support.

Let me try to build one of the x86 arch and see how far it goes.

h-vetinari:
> IMO. we should just keep tensorflow 2.10 if we can't get OSX support.

We're still on an already quite-old grpc version for 2.10. Personally I don't mind dropping osx, but if we don't, we should rebuild 2.10 for grpc 1.51 & openssl 3.

hmaarrfk:
@h-vetinari looking at the OSX logs, it seems that libabseil isn't being used? At least according to conda. Maybe I'm reading the logs badly.

hmaarrfk:
very likely

hmaarrfk:
Ok openssl3 builds pass on OSX x86. I just built them with our private channel enabled, so non-conda forge stuff will appear in the logs. Time to rebuild the 2 i've built then. I'll try to start off some builds tomorrow, but 

6 builds is likely going to take 1-2 days on my donated mac lapotp

recipe/meta.yaml
```diff
@@ -1,20 +1,20 @@
...
       - patches/0001-loosen-requirements.patch
-      - patches/0002-cublas-fix-backport.patch
 
-
-  - url: https://github.com/tensorflow/estimator/archive/refs/tags/v{{ estimator_version.replace(".rc", "-rc") }}.tar.gz
-    sha256: 60df309377cf4e584ca20198f9639beb685d50616395f50770fc0999092d6d85
+  # they didn't release estimator 2.11, but it's needed.. 
+  # - url: https://github.com/tensorflow/estimator/archive/refs/tags/v{{ estimator_version.replace(".rc", "-rc") }}.tar.gz
+  - url: http://github.com/tensorflow/estimator/archive/c7e88f0b110aa1b352e2af3e8876e797a94c27a3.tar.gz
+    sha256: 5275359ff105742016d7a4c0a01b23eb21b30106b11b0b7f27f4cc75bc634726
```
h-vetinari, line 17: Released now: https://github.com/tensorflow/estimator/releases/tag/v2.11.0
```suggestion
  - url: https://github.com/tensorflow/estimator/archive/refs/tags/v{{ estimator_version.replace(".rc", "-rc") }}.tar.gz
    sha256: 922f9187de79e8e7f7d7a5b2d6d3aabc81bbbd6ba5f12a4f52967dd302214a43
```

pkgw:
An error like `bazel-out/host/bin/external/cython/cython_binary: line 4: import: command not found` almost always means that you have a Python program that somehow is being run with `sh` instead. That in turn usually means that there's a missing hashbang `#! /path/to/python` line, or that a program is supposed to be launched as `python3 ./script` and is instead getting launched as `./script` ... which can happen if some kind of configure code is searching for a python3 executable path and not finding anything. In a cross-compile, I'd bet on the latter.

njzjz:
> bazel-out/host/bin/external/cython/cython_binary: line 4: import: command not found

This should be an issue from conda-forge instead of tensorflow. I have the same issue in https://github.com/conda-forge/lammps-feedstock/pull/125, and finally, I add a patch to avoid executing `cython` directly:
```patch
- ${Cythonize_EXECUTABLE} -3 some_code.pyx
 + ${Python_EXECUTABLE} ${Cythonize_EXECUTABLE} -3 some_code.pyx
```

Although, I still don't know how to resolve the original issue.

hmaarrfk:
The problematic file has:
```
#!/Users/mark/git/feedstocks/tensorflow-feedstock/miniforge3/conda-bld/tensorflow-split_1670903470105/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/python

# -*- coding: utf-8 -*-
import re
import sys

from Cython.Compiler.Main import setuptools_main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(setuptools_main())
```
but for cython, it should be taken from the build env. I tried to add cython as a build dep. Gonna start again.

xhochy:
> For reference, this is the error I get if I try to build natively, both using system python and arm64 conda-forge python (to trigger build-locally.py):

That is a `bazel` error. Can you check whether you are using the conda-forge provided bazel or whether some script is secretively downloading a native bazel again?

hmaarrfk:
Slight update. With `cython` added as a build dependency, you end up with the correct shebang line at the top of the script. However, the error remains, so it is likely still trying to execute it as a script, somehow:

<details>

```
#!/Users/mark/git/feedstocks/tensorflow-feedstock/miniforge3/conda-bld/tensorflow-split_1670990536660/_build_env/bin/python

# -*- coding: utf-8 -*-
import re
import sys

from Cython.Compiler.Main import setuptools_main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(setuptools_main()
```
</details>

h-vetinari:
> Btw, @h-vetinari, somewhat unrelated question about openssl3 --- when are we closing the migration and dropping openssl1? Is tensorflow the last big hurdle or are there other big packages we need to address? Somewhat related here because it doubles the manual build process.

See https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/3838. tensorflow is indeed one of the last blockers. It was discussed in the last core meeting - baseline assumption is next spring, but I'd like to get there much sooner. This will depend on getting through the blockers soon and then rediscussing in January.

hmaarrfk:
aren't we still stuck on the arm build for OSX?

pkgw:
I think I figured out the reason for the osx-arm64 error. The problem is that the shebang line in the `cython_binary` script:

```
#!/Users/mark/git/feedstocks/tensorflow-feedstock/miniforge3/conda-bld/tensorflow-split_1670990536660/_build_env/bin/python
```

(from @hmaarrfk's results above) is itself a shell script, basically of the form:

```
#! /bin/bash
exec $PREFIX/bin/python "$@"
```

My testing indicates that on Linux, you can "nest" shebang scripts in this way and the sensible thing will happen, but on OSX it appears that is not the case.

It appears that while `$BUILD_PREFIX/bin/python` is a shell script, `$BUILD_PREFIX/bin/python3` is not, so I think that if we find a way to get the script to embed that path instead, things would work. Or, I think that forcibly overwriting that `python` to be a symlink to `python3` should also work. I'll test the latter approach if I get a chance soon.

hmaarrfk:
Why is the build env python pointing back to the _h_env? is this because of cross python?

hmaarrfk:
https://github.com/conda-forge/cross-python-feedstock/blob/main/recipe/activate-cross-python.sh#L53

pkgw:
Well, the good news is that if I patch the `build.sh` script to remove the custom `$BUILD_PREFIX/bin/python` and make it a symlink to `./python3` (that is, the build-prefix Python binary, not the host-prefix one as cross-python does), I get past the `import` error.

Bad news #1 is that I'm not sure if this will generate correct results. As I understand it, in the cross-python setup, these cross Pythons are special hacked versions that can run on the build architecture but emit metadata for the host architecture, and I think that the version I'm substituting is the non-hacked version that will emit metadata for the host. Especially since this is being used by Cython, I'm worried that this will produce artifacts targeting the wrong arch.

Bad news #2 is that I get a packaging failure later, when trying to create the `tensorflow-base` package. Right after activation I get (after some tidying):

```
$WORK/build_pkg.sh: $PREFIX/bin/pip: $PREFIX/bin/python3.10: bad interpreter: Bad CPU type in executable
```

... because indeed `$PREFIX/bin/python3.10` is an ARM64 executable and I'm running on X86. The `build_pkg.sh` script just does:

```
pip install --no-deps $SRC_DIR/tensorflow_pkg/*.whl
```

During this `tensorflow-core` package build the host directory has Python stuff installed in $PREFIX but nothing in $BUILD_PREFIX, which aligns with the specifications in the `meta.yaml`. I think it might help just to add `python` and `cross-python` to the `build:` section for `tensorflow-core`? But these builds are taking 8 hours so I'd appreciate a second opinion before I fire another one off ...

hmaarrfk:
Maybe you upload a package and ask somebody to install and test it on an M1.

pkgw:
OK, I think that I have a modification to cross-python that addresses https://github.com/conda-forge/cross-python-feedstock/issues/64. At a minimum, with that and a few additional  build deps on `python` and `cross-python`, I can get `osx-arm64` cross-compile build to get up to the `tensorflow-estimator` subpackage.

At that point, the build fails because the build process wants to autogenerate some API wrappers using the script `tensorflow-estimator/tensorflow_estimator/python/estimator/api/create_python_api_wrapper.py`, which eventually wants to import `tensorflow`, which fails basically because we have a version built for the target host architecture (arm64) but can't import its binary modules on the build architecture (x86_64). The specific error I'm getting is about a missing `_pywrap_cpu_feature_guard` binary module, but I suspect that that's just the first symptom of a much bigger general problem.

Previous versions of the Tensorflow package have been successfully built with cross-compiles, right? Have those included tensorflow-estimator? I'm not sure how the existing build scripts could have worked in the cross-compile scenario.

h-vetinari:
> We could have a circular dependency on itself if we want.

That's the whole reason we have `tensorflow-base` package (which `tensorflow-estimator` already depends on), but it seems that this is not enough anymore...?

pkgw:
@ngam I'll post a PR with my proposed cross-python change. It was a bit more of an adventure than I hoped but it does seem to work and I think it solves a real problem.

There is a recent change in the estimator that is relevant, and I can spot at least one bug in it: https://github.com/tensorflow/estimator/commit/0a0444bdf36f89634a141388dff536b6413ba6b2 . However, if I revert that change, it doesn't fix the estimator build :-(

(Not to be coy, the bug is that `flags` is appended to in a function that's called twice, so you get both `--loading=` arguments in the command line.)

pkgw:
My cross-python improvement is PR'd here: https://github.com/conda-forge/cross-python-feedstock/pull/65 . Enjoy the nice clean Rust code!

pkgw:
**If** there's no way to prevent the `tensorflow-estimator` build system from wanting to `import tensorflow`, something that might work — although it would be super annoying — would be to build `tensorflow-base` in its own feedstock, so that a cross-compiled build of the estimator could depend on both "build" and "host" versions of it in its own build. Otherwise, I think that one would have to do the hack that we have for gobject-introspection packages where you build the package twice within `build.sh`, once for build and once for host, and (1) that would take absolutely forever and (2) I'm not sure if it would even be possible to do across a subpackage split. Or, we just give up on cross-compiling the package. But all of this is only necessary if we can't fix the cross-compile here and I'm still hopeful that we can figure out how to do that.

hmaarrfk:
If you try that, and it works, i can start to uoad the osx builds for you on conda forge loosely following CFEP03...

pkgw:
Note for anyone attempting builds: as of ~today you have to add a constraint for `bazel < 6` in the appropriate places.

h-vetinari:
> Also, is it okay to drop the openssl1 variants or should I build them as well?

As long as the migration isn't finished I'd say by default both variants, but given the effort involved here, if it's too much to handle, just the openssl 3 builds would suffice. A preview of closing the migration, so to speak. ;)

pkgw:
@ngam Great to hear that you had a successful build!

The tensorflow-estimator build process now (?) includes a step that needs to import the Tensorflow Python modules in order to generate some set of API bindings. I believe that with the current build files, a Rosetta build runs ARM64 Python and everything is OK, but a proper cross build will have cross-python as a build dep for the `tensorflow-estimator` subpackage, and in that case the import fails, leading to (after some tidying):

```
INFO: Analyzed target //tensorflow_estimator/tools/pip_package:build_pip_package (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
[0 / 1] [Prepa] BazelWorkspaceStatusAction stable-status.txt
SUBCOMMAND: # //tensorflow_estimator:estimator_python_api_gen_compat_v1 [
  action 'Executing genrule //tensorflow_estimator:estimator_python_api_gen_compat_v1', 
  configuration: c75b52840eea6dcbe1b8559af2885f9d37de38614db15e782e3b011d421f690b, 
  execution platform: @local_config_platform//:host
]
(cd $BUILD_PREFIX/share/bazel/4f69aca1870d66a6fb89c3936195a068/execroot/org_tensorflow_estimator && \
  exec env - \
    PATH=$BUILD_PREFIX/bin:$PREFIX/bin:$BUILD_PREFIX/bin:$PREFIX/bin:/private/tmp/zztmp/bin:$PREFIX/bin:/Users/peter/miniforge3/condabin:/Users/peter/.repo/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/Library/Apple/usr/bin \
  /bin/bash -c '\
    source external/bazel_tools/tools/genrule/genrule-setup.sh; 
    bazel-out/darwin-opt-exec-2B5CBBC6/bin/tensorflow_estimator/create_tensorflow_estimator.python.estimator_api_1_estimator_python_api_gen_compat_v1  \
      --apidir=bazel-out/darwin-fastbuild/bin/tensorflow_estimator_api/v1/ \
      --apiname=estimator \
      --apiversion=1   \
      --packages=tensorflow_estimator.python.estimator \
      --output_package=tensorflow_estimator._api.v1 \
      --loading=static \
      --loading=default \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/v1.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/__init__.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/experimental/__init__.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/export/__init__.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/inputs/__init__.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/tpu/__init__.py \
      bazel-out/darwin-fastbuild/bin/tensorflow_estimator/_api/v1/estimator/tpu/experimental/__init__.py'
)

# Configuration: c75b52840eea6dcbe1b8559af2885f9d37de38614db15e782e3b011d421f690b
# Execution platform: @local_config_platform//:host

ERROR: $WORK/tensorflow-estimator/tensorflow_estimator/BUILD:74:19: \
  Executing genrule //tensorflow_estimator:estimator_python_api_gen_compat_v1 failed: (Exit 1): sandbox-exec failed: error executing command 
  [...]

  File "$BUILD_PREFIX/share/bazel/4f69aca1870d66a6fb89c3936195a068/sandbox/darwin-sandbox/31/execroot/org_tensorflow_estimator/bazel-out/darwin-opt-exec-2B5CBBC6/bin/tensorflow_estimator/create_tensorflow_estimator.python.estimator_api_1_estimator_python_api_gen_compat_v1.runfiles/org_tensorflow_estimator/tensorflow_estimator/python/estimator/api/create_python_api_wrapper.py", line 26, in <module>
    from tensorflow_estimator.python.estimator import estimator_lib  # pylint: disable=unused-import
  File "$BUILD_PREFIX/share/bazel/4f69aca1870d66a6fb89c3936195a068/sandbox/darwin-sandbox/31/execroot/org_tensorflow_estimator/bazel-out/darwin-opt-exec-2B5CBBC6/bin/tensorflow_estimator/create_tensorflow_estimator.python.estimator_api_1_estimator_python_api_gen_compat_v1.runfiles/org_tensorflow_estimator/tensorflow_estimator/python/estimator/estimator_lib.py", line 22, in <module>
    from tensorflow_estimator.python.estimator.canned.baseline import BaselineClassifier
  File "$BUILD_PREFIX/share/bazel/4f69aca1870d66a6fb89c3936195a068/sandbox/darwin-sandbox/31/execroot/org_tensorflow_estimator/bazel-out/darwin-opt-exec-2B5CBBC6/bin/tensorflow_estimator/create_tensorflow_estimator.python.estimator_api_1_estimator_python_api_gen_compat_v1.runfiles/org_tensorflow_estimator/tensorflow_estimator/python/estimator/canned/baseline.py", line 53, in <module>
    import tensorflow as tf
  File "$BUILD_PREFIX/venv/lib/python3.10/site-packages/tensorflow/__init__.py", line 37, in <module>
    from tensorflow.python.tools import module_util as _module_util
  File "$BUILD_PREFIX/venv/lib/python3.10/site-packages/tensorflow/python/__init__.py", line 36, in <module>
    from tensorflow.python import pywrap_tensorflow as _pywrap_tensorflow
  File "$BUILD_PREFIX/venv/lib/python3.10/site-packages/tensorflow/python/pywrap_tensorflow.py", line 26, in <module>
    self_check.preload_check()
  File "$BUILD_PREFIX/venv/lib/python3.10/site-packages/tensorflow/python/platform/self_check.py", line 63, in preload_check
    from tensorflow.python.platform import _pywrap_cpu_feature_guard
ImportError: cannot import name '_pywrap_cpu_feature_guard' from 'tensorflow.python.platform' ($BUILD_PREFIX/venv/lib/python3.10/site-packages/tensorflow/python/platform/__init__.py)
```

The script that is run during the build process is [`tensorflow_estimator/python/estimator/api/create_python_api_wrapper.py`][1].

[1]: https://github.com/tensorflow/estimator/blob/master/tensorflow_estimator/python/estimator/api/create_python_api_wrapper.py

pkgw:
OK, I think I've figured out the modifications necessary to cross-compile for osx-arm64 on a non-Rosetta, x86_64 host. I think they make the build approach more formally "correct" than what we currently have, but would break some workflows. So @ngam if you are OK with doing all of the macOS builds I think it would be best for me to propose those changes in a separate PR, once this update is worked out.

@ngam I think the difference is that my machine really truly can't run arm64 executables. So, when trying to build `tensorflow-base` with this PR as-is, for instance, I get a `Bad CPU type in executable` error, because `build_pkg.sh` does a `pip install` which invokes `$PREFIX/bin/python` which is the target host (ARM64) version — if you look at the build logs, the build environment for that step (`$BUILD_PREFIX`) doesn't have Python installed at all. Your setup runs the exact same code, but while it "shouldn't" be able to run the target host Python it actually can. The "more correct" approach is to install `python` and `cross-python_{blah}` in the build environment when `build_platform != target_platform`, which sets up a very specially crafted environment that turns `$PREFIX/bin/python` into something that will actually run on the build host arch. But then you run into the `import tensorflow` issue in the estimator build, due to the characteristics of the specially-crafted environment.

hmaarrfk:
I didn't build all of them, just a proof of concept one. 

I am not next to a computer with a stable internet connection at the moment.

pkgw, you can open a PR onto this branch if you feel like showing us your changes in the meantime.

pkgw:
@hmaarrfk My fixes are on [pkgw/tensorflow-feedstock:tf211][1]. It's only a few changes, but adding a build dep of `tensorflow-base` to `tensorflow-estimator` is going to break `build-locally.py` builds, I think.

[1]: https://github.com/pkgw/tensorflow-feedstock/tree/tf211

h-vetinari:
> adding a build dep of `tensorflow-base` to `tensorflow-estimator` is going to break `build-locally.py` builds, I think.

Shouldn't that be fine if the osx-64 builds are available at that time?

hmaarrfk:
> to break build-locally.py builds, I think.

I think you might be correct. The osx buikd locally works a little differently than the Linux one.

But we could likely start to upload the normal builds, upload them to conda forge, and then work on the cross compiled builds.

pkgw:
I was going to work on generating more osx-arm64 builds and uploading them, but it turns out that my Anaconda.org account has a small quota and will absolutely not be able to host all of the package files! If I upload them to some kind of other storage, would there be a way to work around that?

hmaarrfk:
I can start uploading packages to conda-forge if you point me to where they are now. I've created a checkecklist in the first post so that we can keep track of things.

pkgw:
@hmaarrfk My upload spot is https://anaconda.org/pkgw-forge/repo . I've uploaded OpenSSL3 / Python 3.10 × {osx-64, osx-arm64} right now (since I need to do the non-cross build to complete the cross build). I'll try to work through the other OpenSSL3 variants this week, although each individual build has to run overnight for me.

akrherz:
> I uploaded all the linux ones, @hmaarrfk. Did you check these?

@ngam I attempted to check out the linux ones, but notice no linux 64 for the [tensorflow package](https://anaconda.org/ngam/tensorflow/files)

pkgw:
@ngam Thank you for testing, that's awesome to hear! I still need to test one final piece of the feedstock changes; it looks like I need to deselect most of the `test: commands:` bits that won't work in a cross-build scenario, and haven't actually finished a build with the latest version of my changes there.

pkgw:
@hmaarrfk I've just uploaded packages for Python 3.8. A couple more days and hopefully I'll get the Python 3.9 builds done and finish off the macOS builds! (I am intending to go with the suggestion above and not build OpenSSL 1.1 versions.)

pkgw:
My anaconda account `pkgw-forge` now contains all six macOS builds: OpenSSL3 × { Python 3.8, 3.9, 3.10 } × { osx-64, osx-arm64 }. Here are the build logs:

[logs.zip](https://github.com/conda-forge/tensorflow-feedstock/files/10369219/logs.zip)

I never got the cross-builds to pass the testing phase due to cross-execution issues but that doesn't mean that there's anything wrong with the builds. (I think I must be doing something wrong with setting up the cross build invocation ... my last error was with the `imports:` test step which mambabuild must be smart enough to realize won't work on a cross environment, no?)

hmaarrfk:
the build locally command should skip the tests right? 

Sorry i just got back from travels, i'll move the builds over.

hmaarrfk:
I've tested python 3.9 and cuda, as well as CPU on linux using my workflow. seems to work ok. i'm going to upload all. Thanks for the hard work.

I seemed to have a problem with openssl3+cuda on linux though. I can't seem to point it down.

hmaarrfk:
Ok, all got uploaded. Thanks all. Onto the next stuff.

Seems like grpc 1.51 also got folded in.

h-vetinari:
Woohoo, thanks everyone for all the hard work! 🥳 

> Seems like grpc 1.51 also got folded in.

Yes, that was explicitly a goal.

hmaarrfk:
> Yes, that was explicitly a goal.

It really became hard to follow what got "done" especially since the migration file got deleted.

Ultimately, i checked a few logs.

h-vetinari:
> It really became hard to follow what got "done" especially since the migration file got deleted.

That's because the migration got closed in the meantime (it was clear that the next TF build would contain it), and it's right there in the diff (for the config yamls). But happy to answer any questions about these things next time.

hmaarrfk:
i think what might help as these CFEP03 builds get done is if somebody wanted to keep the top level comment updated with highlights such as this one.

Thanks for help on all of these difficult builds!

PR #105, opened 2020-07-09
1.14.0

brendanf:
Checklist
* [X] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [X] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [X] Ensured the license file is being packaged.

Although I see that work is ongoing for even newer versions of TensorFlow, I am trying to build a package that specifically requires 1.14, so I've carried on with some modifications to hopefully get #93 working on all platforms.

PR #297, opened 2023-01-16
Updating meta.yaml for 1.15.0

holtgrewe:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

This version is required for the Python packages for the popular bioinformatics package GATK4.


holtgrewe:

h-vetinari:
Some immediate thoughts:
* 1.15.0 is [over](https://github.com/tensorflow/tensorflow/tags?after=v2.1.0-rc1) 3 years old
* There's also a [1.15.5](https://github.com/tensorflow/tensorflow/tags?after=v2.0.4) (though also 2 years old already), so if anything we should target that
* this is targeting the wrong branch (we'd have to create one for pre-2.0, which was the state before https://github.com/conda-forge/tensorflow-feedstock/pull/110)

> This version is required for the Python packages for the popular bioinformatics package GATK4.

I can create a branch for you, but since this feedstock regularly times out when building anything, I doubt you'll attract much interest in volunteers to build those packages (unless you're ready/willing to do that yourself) -- it's a lot of work already as-is, and I fear that building long-obsolete packages is not something that will achieve much momentum.

holtgrewe:
@ngam @hmaarrfk Hello. I would like to see this (super old) version in conda-forge. Am I doing this right?

dlaehnemann:
@h-vetinari Probably the state after the merge of the update to `1.14.0` is the best to base a new branch on. Could you create that? Or could we do this ourselves, without any special permission on the feedstock?
This is the PR: https://github.com/conda-forge/tensorflow-feedstock/pull/105
Here's the commit: https://github.com/conda-forge/tensorflow-feedstock/commit/2342fc2f4fc64378f794316c2f09ce8847f484b3

Generally, such a branch is the way to go, right?
https://conda-forge.org/docs/maintainer/updating_pkgs.html#maintaining-several-versions

And generally, I would second @holtgrewe that the `bioconda` package GATK4 that needs this particular version (`1.15.0`) is very important in bioinformatics. However, neither of us is a developer of the tool itself, so we can't fix its dependencies over there. But with (hopefully very little) help, we can maybe fix it by building a conda-forge package for this old tensorflow version...

h-vetinari:
OK, I pushed a branch and pointed this PR to it. You'll need to revert d76d41ab373d15d18ef32d18c18023bd33e64147, then rebase your changes on top, then rerender.

Note that the old approach of binary repackaging of wheels is not going to be acceptable anymore (speaking only for me personally, especially with all the problems about `_GLIBCXX_USE_CXX11_ABI` and related crashes), so you should try to recompile this from source with similar changes as was done in #110

dlaehnemann:
Many thanks already, this should get us going.

And yes, I also thought `1.15.5` should be an option, and shouldn't contain anything incompatible with `1.15.0`. But then you never know. So I'd go for `1.15.0` first, and maybe `1.15.5` in a second step off of that.

h-vetinari:
No problem.

You'll very likely need to keep most of the patches that were being used here for 1.14. To rebase them, I suggest following something along the [following](https://github.com/conda-forge/ray-packages-feedstock/blob/main/ray-patching.md) lines (it's written for a specific feedstock, but is really generally applicable to any feedstock carrying patches longterm).

dlaehnemann:
Sounds like fun...

Thanks already for all the pointers, we'll try to pick this up later this week.

hmaarrfk:
So, i don't mean to discourage you, but I do think that this is a place where `pip` shines.

`conda` tries to ensure that shared C libraries are available to different packages. This helps reduce the maintenance burden to somebody trying to use a C library. They don't need to compile it themselves. They can use the one installed by conda and link to it.

`pip` takes the other approach. Every C library you need, except for the python standard library, and some old version of the C standard library, you need to include as part of your package. What this means is that there is more maintenance burden on the packagers side, but from the end user's side, it is somewhat "easier to use" as a python library. 

Because of this, I think you will have more luck:

* Finding a version of numpy that works for 1.15
* Building most of your environment with conda
* Installing what you need from tensorflow with pip

Otherwise, you will be stuck trying to compile "old tensorflow" with "new conda-forge".

If you do try to attempt to build a conda package, I would:

1. Focus on the versions of python you need. Looking at their releases, they only released wheels for 3.6 and 3.7. Both are versions of python that conda-forge has dropped support for. https://pypi.org/project/tensorflow/1.15.5/#files  you can try to use 3.8 but I remember there was some kind of hurdle on the C side (even if minor). The patches we have might help.
2. Focus on the operating systems you need. Do you need OSX + linux? or just OSX? 
3. Focus on the architectures you need. Anything other than x86 is likely going to fail.
4. Focus on the cuda builds you need. 

Best of luck. I gave up trying to do the same thing for 2.9.X a few weeks back, but maybe if you constrain your problem, you can get somewhere.

hmaarrfk:
can we close this?

PR #343, opened 2023-10-12
Effectively disable CI on this feedstock

xhochy:
The CI jobs for the full matrix always timeout after six hours. As the matrix is quite large, disable it so that it doesn't block other jobs from running.

### DIFF
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -1,3 +1,8 @@
+azure:
+  settings_linux:
+    timeoutInMinutes: 1
+  settings_osx:
+    timeoutInMinutes: 1
 build_platform:
   osx_arm64: osx_64
 conda_build:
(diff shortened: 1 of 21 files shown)


conda-forge.yml
```diff
@@ -1,3 +1,8 @@
+azure:
+  settings_linux:
+    timeoutInMinutes: 1
+  settings_osx:
+    timeoutInMinutes: 1
```
hmaarrfk, line 5: Amazing!

shall we revert the pinning changes?

xhochy: I would like to do this in a separate PR. They make take some time due to `nvcc` weirdness and I simply want to minimise the load impact our work on this feedstock has.

hmaarrfk: i meant

```
git checkout main -- .ci_support
```
to undo the changes in the ci_support files that relate to pinning upgrades that didn't happen.

xhochy: Ah, got it!

PR #398, opened 2024-07-16
Bump 2.17.0

xhochy:
GitHub doesn't let me push to #394 anymroe


h-vetinari:
We may want to include the icu migration as well...?

xhochy:
Let's first get this to green. I had some problems with `icu` on other feedstocks, so I'm eager not to add more things to this PR until it suceeds.

PR #438, opened 2025-08-01
Add linux-aarch64 and cuda 12.8

dslarm:
This PR is a cleaned and updated branch to bring in both aarch64 and also cuda 12.8. This replaces the PR #426.

Nutshell of #426 learnings:
* cross-compilation was a fool's errand - there are too many issues to resolve, we must therefore use native builds.
* cuda support requires version 12.8 - as cuda 12.6 can't handle Arm NEON - and even 12.8 requires pinning to an earlier gcc (11)

Recent learnings:
* gcc 14 will error where gcc 13 was happy to warn - we need to pin to gcc <= 13 across all platforms

Next
* reviewers: please review

Outstanding questions - may or may not need me to do something:
* what conda-forge.yml should be - I know I will need to native-build the binaries for a while, we're still waiting native linux-aarch64 builds via CI
* and just curious.. how CI works for this monster.. it's 45 mins on a 192 core x86 or aarch64 box per combo..


hmaarrfk:
Thank you for troubleshooting some GCC + CUDA incompatibilities.


> reviewers: please review

I simply cannot review anything that doesn't address the broken status of this recipe for other platforms.

Conda-forge is where it is today because we strive for compatibiliy with many platforms.

x86 on linux and mac builds must be addressed first, before I can spend time thinking about aarch64.

My recommendation remains the same:
- Create your own channel
- tell users to add it, in addition to conda-forge's
- Upload your packages there while we have time to address the other breakages.

hmaarrfk:
rebasing might be hard for the build script. i agree that it might be easier to just "redo" the changes to that one manually.

dslarm:

dslarm:
> This is ready for review:
> 
Anyone able to review this?  Thanks.

h-vetinari:
> Anyone able to review this? Thanks.

I'm not super excited about having to build aarch separately. I seem to remember that somewhere along your work it came up that cross-compilation had some issues? Do you have a reference somewhere what those were?

It would be very beneficial IMO if aarch+CUDA could be built on the opengpu server (which would require cross-compilation though), because the alternative is that we'll have to ping you (or whoever else) to locally build linux-aarch64 + CUDA according to CFEP-03

dslarm:
@h-vetinari - thanks for your feedback, and patience - I believe I've finally nailed cross-compile, found something I'd missed last time and we can now reach the end.  The downside is that means disabling a lot of tests (python imports for example, failed as running python has wrong arch).

So - we now have with this from this PR:
1. cuda 12.8 adds for all linux
2. cross-compile linux-aarch64 - cuda and non-cuda

h-vetinari:
@dslarm, unfortunately your odyssey isn't finished yet. There was no CUDA 12.8 job for aarch here yet, and now that I've added it, it runs into:
```
 [13 / 26] [Prepa] BazelWorkspaceStatusAction stable-status.txt
ERROR: /home/conda/feedstock_root/build_artifacts/tensorflow-split_1757278331355/_build_env/share/bazel/a5889192f1201e14ae645981f8e2d4ca/external/double_conversion/BUILD:9:11: Compiling double-conversion/strtod.cc [for tool] failed: undeclared inclusion(s) in rule '@double_conversion//:double-conversion':
this rule is missing dependency declarations for the following files included by 'double-conversion/strtod.cc':
```

h-vetinari:
Sigh
```
 The following packages are incompatible
├─ tensorflow-avx2 =2.18.0 cpu_py313hf8d5db8_51 is not installable because it requires
│  └─ tensorflow ==2.18.0 cpu_py313h1234567_51, which does not exist (perhaps a missing channel);
                                   ^^^^^^^^
                                   dummy hash
```
reminds me of https://github.com/conda/conda-build/issues/5571

hmaarrfk:
does it pass with aarch native compilation dslarm was willing to do that and invoke CFEP03.

I'm personally too burned out to care about anything that doesn't "unpin" abseil on tensorflow.

Its causing me real solving issues, where alot of the stack is getting downgraded to add tensorflow to an environment.

